PR #8, opened 2022-09-01
Fix build

kmuehlbauer:
Testing if bumbversion-section in setup.cfg interferes with setuptools_scm


kmuehlbauer:
Another thing to check triggering "release:created" vs "release:published"

kmuehlbauer:
@mgrover1 I've added a workflow to just test the creation of sdist/wheel for a supposed upload to pypi and testpypi. This does not do any uploading. It seems that the packages get created with the correct numbering scheme.

mgrover1:
In PRs, this will not have access to the event id... I would suggest running tests in the CI instead of uploading to pypi

kmuehlbauer:
@mgrover1 I try to test, if the two different build schemes are working and uploading the artifact to gha space. Not trying to check this on PyPI.

mgrover1:
Ahh - okay!

PR #12, opened 2022-09-05
Fix version 999 issue

kmuehlbauer:


kmuehlbauer:
This fixes the issue with 999 version on RTD. Moving forward.

PR #13, opened 2022-09-10
Data model + CfRadial1 Backend

kmuehlbauer:
First code porting from wradlib:

- data model 
- CfRadial1 Backend

There is a first import from CfRadial1 into the DataTree.


mgrover1 (changes requested):
This looks great @kmuehlbauer ! The only suggestion I would have is to add a test for the cfradial io.

kmuehlbauer:
@mgrover1 Do we have a cfradial1 file somewhere? I have one inside wradlib-data repo. But that would be overkill. Shall we create a xradar-data repo where we host our testfiles?

mgrover1:
@kmuehlbauer - great question. I think a more general open-radar-data repo would be great. This way, we can share across repositories and such.

mgrover1:
I also do not see an issue in using one of the wradlib-data files, or one of the Py-ART ones [used here](https://arm-doe.github.io/pyart/examples/plotting/plot_ppi_cfradial.html#sphx-glr-examples-plotting-plot-ppi-cfradial-py)

kmuehlbauer:
I'll just fetch the one file from wradlib-data repo and see how it goes. For CI no problem since we need internet connection anyway.

kmuehlbauer:
@mgrover1 I've added two tests, one for open_cfradial_datatree and one for xr.open_dataset. Added fixture to grab the cfradial1 file from wradlib-data repo. 

This just works with the minimal required data. We would need to enhance this to also make optional attributes and variables available.

kmuehlbauer:
@mgrover1 This is good to go from my end. There is still much to do, though.

PR #20, opened 2022-09-16
DOC: add importers.md, documentation on importers

kmuehlbauer:


mgrover1 (approved):
This make sense - thanks @kmuehlbauer ! In a future PR, I think we should show how users can use the datatree io with specifying which backend to use too!

PR #23, opened 2022-09-20
Fix version on ReadTheDocs

kmuehlbauer:
For some reason the tagged versions (eg. 0.0.6) are not correctly installed. This version shows up as `0.0.7.dev0+g8f2753a.d20220920`.

This PR checks the commit on RTD and pip installs that commit directly from github. 

~This might break for PR builds, which we would need to detect and prepare for.~

This looks like it works nicely also for PR builds.


kmuehlbauer:
@mgrover1 This should work also for releases. Please merge at your discretion and release 0.0.7 to be sure.

mgrover1 (approved):

kmuehlbauer:
@mgrover1 OK, let's get this in.

PR #28, opened 2022-09-26
ENH: Add georeference module, basic tests

mgrover1:
Add some georeference helper functions, first steps to georeferencing the xradar object


kmuehlbauer:
Thanks @mgrover for setting this up. :+1: 

Just some questions, comments and thoughts:

- Is there a specific need to have `geographic_to_cartesian_aeqd` in the codebase?
- It would be great to parameterize earth radius as well as k.
- I'd suggest to use metric SI units (meters in this example), as the model keeps the ranges in meters.
- We can use site-coordinates (lat, lon, alt) to calculate final bin aeqd coordinates.
- We might use latitude dependend earth radius for the calculations.
- We might return a projstring with the used AEQD projection.

Basically in wradlib we are using the same equations derived Doviak. We might utilize the existing wradlib function which adds the x, y, z and groundrange coordinate DataArrays.

mgrover1:
Yeah - on second thought, I like that idea of operating at the **dataset level** instead of the numpy level. This is using the current functions/methods we have in Py-ART, but adapting it to work at the dataset level is a better approach.

I will make those changes, and ping you when it is ready for review.

xradar/georeference/transforms.py
```diff
@@ -0,0 +1,116 @@
...
+    elevation angle, s is the arc length, and R is the effective radius
+    of the earth, taken to be 4/3 the mean radius of earth (6371 km).
+    References
+    ----------
+    .. [1] Doviak and Zrnic, Doppler Radar and Weather Observations, Second
+        Edition, 1993, p. 21.
+    """
+    if effective_radius_fraction is None:
+        effective_radius_fraction = 4.0 / 3.0
+    theta_e = np.deg2rad(elevations)  # elevation angle in radians.
```
kmuehlbauer, line 50: Just a personal style thing, I'd prefer comments above. But no strict necessity.

xradar/georeference/transforms.py
```diff
@@ -0,0 +1,116 @@
...
+    ds['x'], ds['y'], ds['z'] = antenna_to_cartesian(
+        ds.range,
+        ds.azimuth,
+        ds.elevation,
+        earth_radius=earth_radius,
+        effective_radius_fraction=None,
+    )
+
+    # Set the attributes for the dataset
+    ds.x.attrs = {'standard_name': 'east_west_distance_from_radar', 'units': 'meters'}
```
kmuehlbauer, line 107: Let's move those attributes into the model.

examples/notebooks/plot-ppi.ipynb
```diff
@@ -0,0 +1,151 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "eaa2ee4c-98f2-41d2-b16e-9d05ad8608da",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ds = xd.georeference.get_x_y_z(ds)\n",
```
kmuehlbauer, line 98: Maybe also advertise the `.pipe()`

`ds = ds.pipe(georeference.get_x_y_z)`

tests/georeference/test_transforms.py
```diff
@@ -0,0 +1,56 @@
...
+    assert_almost_equal(y[np.where(azimuths == 270.0)], 0)
+
+
+def test_get_x_y_z():
+
+    # Create an empty xradar dataset
+    ds = xradar.model.Dataset()
+
+    # Add similar data to before
+    ds['time'] = pd.date_range(start='2018-11-10', end='2018-11-10T00:09:00', freq='T')
```
kmuehlbauer, line 35: We could facilitate `xradar.model.create_sweep_dataset` here.

kmuehlbauer (approved):
@mgrover1 This looks very neat throughout. Just a couple of minor comments and suggestions. Those might be addressed here on a follow up PR.

I've fixed a bit for the docs, to be able to see it in all it's glory.

tests/georeference/test_transforms.py
kmuehlbauer: Double take here

tests/georeference/test_transforms.py
```diff
@@ -0,0 +1,56 @@
...
+    assert_almost_equal(x[np.where(azimuths == 180.0)], 0)
+
+    # Make sure that at 270 degrees, y is (close to) 0
+    assert_almost_equal(y[np.where(azimuths == 270.0)], 0)
+
+
+def test_get_x_y_z():
+
+    # Create an empty xradar dataset
+    ds = xradar.model.Dataset()
```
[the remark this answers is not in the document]
mgrover1: I like these approaches! Let's add these to the examples too.

xradar/georeference/transforms.py
mgrover1: Me too - I'll correct that

examples/notebooks/plot-ppi.ipynb
mgrover1: Yeah - that might be good - I think this would be good in a follow up PR

xradar/georeference/transforms.py
mgrover1: > @mgrover1 This looks very neat throughout. Just a couple of minor comments and suggestions. Those might be addressed here on a follow up PR.

Can we address these with a follow-up PR?

mgrover1:
> 
Let's address these with another PR.

kmuehlbauer:
Ready to go, @mgrover1 ?

mgrover1:
Sure :)

kmuehlbauer:
Thanks @mgrover1, this is going to be fun to add to the current examples.

PR #33, opened 2022-09-28
ENH: Iris/Sigmet importer

kmuehlbauer:


kmuehlbauer:
That's the last importer for ray-based data. There are some more cartesian based ones available, which we might consider later in case the standard moves into that direction.

kmuehlbauer:
Final cleanup. This is good to go from my end.

mgrover1 (approved):
Great! Let's merge!

PR #36, opened 2022-09-29
XYZ shape

kmuehlbauer:
Some broadcasting tweaks to get the xyz coords into (time, range).

It might be good to have them in the same shape as the data arrays.


kmuehlbauer:
Please merge at your discretion. I'm getting back later today. Gotta run now.

PR #39, opened 2022-10-02
ENH: add odim exporter, add roundtrip test, add minimal documentation

kmuehlbauer:


mgrover1:
@kmuehlbauer just let me know when this is ready for review

kmuehlbauer:
@mgrover1 I think it is read for a first review. I've stripped this to a minimum.

xradar/io/export.py
```diff
@@ -0,0 +1,264 @@
...
+
+import datetime as dt
+
+import h5py
+import numpy as np
+
+from ..model import required_sweep_metadata_vars
+
+
+def _write_odim(src, dest):
```
mgrover1, line 42: Can you write out source and destination here?

mgrover1 (commented):
One minor comment - otherwise, this looks great! Can we add some sample odim_h5 data to that radar data repo? And use it here as a verification of standards?

xradar/io/export.py
kmuehlbauer: @mgrover1 Yes, Let me add this here

kmuehlbauer: done

mgrover1 (approved):

mgrover1 (changes requested):
One small linting error

mgrover1 (approved):

kmuehlbauer:
Finally, had some old versions of isort/black in my environment. Should work now.

kmuehlbauer:
Imagine a life without Cloud CI ... :grimacing:

PR #44, opened 2022-11-10
MNT: fetch radar data from open-radar-data, add description/url

kmuehlbauer:
- fetch radar data from open-radar-data
- add description/url to BackendEntrypoints
- test notebooks, fixup CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,11 @@ jobs:
     name: lint and style checks
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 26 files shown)


kmuehlbauer:
@mgrover1 Just some maintenance on code and CI.

mgrover1 (approved):

mgrover1:
@kmuehlbauer I think we can go ahead and merge. For some reason code coverage is dropping, but still above 80% which is good!

kmuehlbauer:
> @kmuehlbauer I think we can go ahead and merge. For some reason code coverage is dropping, but still above 80% which is good!

There are more lines removed which counted to coverage before than added, so the overall coverage percentage decreased a bit.

PR #43, opened 2022-11-09
Make cfradial generic

mgrover1:
Closes #42

PR #45, opened 2022-11-12
ENH: Generalize CfRadial1 reader

kmuehlbauer:
This is an attempt to increase the amount of metadata which is imported into the internal DataTree-structure at the right places CfRadial2 is expecting it.

Takes over from: #43 

- refactor internal cfradial1 reader to extract CfRadial2 groups from cfradial1 Dataset
- enhance the model with subgroup metadata
- fix tests
- add Notebook showing the transformation from CfRadial1 to CfRadial2
- fix Notebooks and tests

ToDo:

- Make the above happen also for sweep-subgroups (need to have example files).
- Make the above happen also for other sources (`ODIM_H5`, etc), the transformation might not be that easy, though. But in any case a notebook writeup showing the transformation would help developers and users to look behind the scenes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
+    rev: v4.3.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -24,6 +24,7 @@ notebooks/Accessors
 :maxdepth: 2
 :caption: Examples
 
+notebooks/CfRadial1_Model_Transformation
 notebooks/CfRadial1
 notebooks/CfRadial1_full
 notebooks/ODIM_H5
--- a/examples/notebooks/Accessors.ipynb
+++ b/examples/notebooks/Accessors.ipynb
@@ -61,9 +61,7 @@
   {
    "cell_type": "markdown",
    "id": "0ed04dfa-eccc-4ca6-8864-25ce467518e2",
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "source": [
     "### Open data"
    ]
@@ -292,7 +290,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.13"
+   "version": "3.10.6"
   }
  },
  "nbformat": 4,
--- a/examples/notebooks/CfRadial1.ipynb
+++ b/examples/notebooks/CfRadial1.ipynb
@@ -3,9 +3,7 @@
   {
    "cell_type": "markdown",
    "id": "59447ad6-ac47-494e-b696-4335b36b205b",
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "source": [
     "# CfRadial1 - Simple"
    ]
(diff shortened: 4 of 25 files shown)


kmuehlbauer:
@mgrover1 I've squashed your commits from #43 into one and added my changes on top of it. 

The initial implementation wasn't that bad already :-)

kmuehlbauer:
@mgrover1 It would be great, if you could test this PR against some of your CfRadial1 files if you find the time. 

Dual `frequency` cases would be great to have and also sweep subgroup cases would be interesting (also for testing). So if anyone has some of these special case CfRadial1 data files available for open sharing, that would be great.

mgrover1:
This looks great @kmuehlbauer - I tried it on the sample dataset, and it worked well

```python
print(ds['sweep_0'])

DataTree('sweep_0', parent="root")
    Dimensions:                      (azimuth: 1154, range: 675)
    Coordinates:
        time                         (azimuth) datetime64[ns] 2022-08-25T19:47:14...
      * range                        (range) float32 -118.4 -58.48 ... 4.028e+04
      * azimuth                      (azimuth) float32 0.002944 0.07436 ... 359.9
        elevation                    (azimuth) float32 1.011 1.011 ... 1.011 1.011
        latitude                     float64 38.9
        longitude                    float64 -106.9
        altitude                     float64 3.149e+03
    Data variables: (12/28)
        DBZ                          (azimuth, range) float64 ...
        VEL                          (azimuth, range) float64 ...
        WIDTH                        (azimuth, range) float64 ...
        ZDR                          (azimuth, range) float64 ...
        PHIDP                        (azimuth, range) float64 ...
        RHOHV                        (azimuth, range) float64 ...
        ...                           ...
        time_coverage_end            |S192 b'2022-08-25T19:47:14Z'
        time_reference               |S192 b'1970-01-01T00:00:00Z'
        volume_number                int32 218053
        platform_type                |S192 b'fixed'
        instrument_type              |S192 b'radar'
        primary_axis                 |S192 b'axis_z'
```

mgrover1 (approved):

kmuehlbauer:
@mgrover1 This looks like the version you run is not the correct version. I've found that installing xradar via pip sometimes keeps old versions (and uses them). I did not come behind that. I'm removing those by hand from site-packages-folder.

In your case the `latitude`/`longitude`/`altitude` should not be in the sweep-Dataset. They can be found in the root-group. 

If the user reads a complete volume as DataTree they will get site-coords with the root-group. But if they open a single sweep with `xr.open_dataset` they will not get the site-coords with the sweep. I'm wondering if we should allow site-coordinates in the sweep-Dataset when loaded with the xarray-Backend? Or would that confuse users, if they don't have the site-coords in the sweep-groups (`DataTree`) but have them in the sweep-Datasets (`xr.open_dataset`)?

If this is something we can defer to later discussion, I'm good with merging this. I'll open follow-up issues for the todo's listed above.

kmuehlbauer:
Here is sweep_0-group with the correct reader with this PR.

```
DataTree('sweep_0', parent="root")
    Dimensions:                    (time: 483, range: 996)
    Coordinates:
      * time                       (time) datetime64[ns] 2008-06-04T00:15:03 ... ...
      * range                      (range) float32 150.0 300.0 ... 1.494e+05
        azimuth                    (time) float32 ...
        elevation                  (time) float32 ...
    Data variables: (12/18)
        sweep_number               int32 ...
        sweep_mode                 |S32 ...
        prt_mode                   |S32 ...
        follow_mode                |S32 ...
        sweep_fixed_angle          float32 ...
        pulse_width                (time) timedelta64[ns] ...
        ...                         ...
        r_calib_index              (time) int8 ...
        measured_transmit_power_h  (time) float32 ...
        measured_transmit_power_v  (time) float32 ...
        scan_rate                  (time) float32 ...
        DBZ                        (time, range) float32 ...
        VR                         (time, range) float32 ...
```

mgrover1 (approved):

kmuehlbauer:
OK, so hopefully everyone is aware of the above minor issues. Pulling this in.

Issue #48, opened 2022-11-15
Site-coordinates at Sweep-Level


kmuehlbauer:
In CfRadial2-standard the sweep-groups do not contain the site-coordinates (`latitude`/`longitude`/`altitude`). Although that makes perfectly sense when reading the whole DataTree, it might be not as perfect for the single-sweep reading via `xr.open_dataset`. The site-coordinates (at least `latitude` and `altitude`) might be used for georeferencing.

I'd suggest to enable the xarray-backends to add the site-coordinates to the returned Dataset. We have at least the following options:

- return site-coords per default in any case
- return site-coords per kwarg-request per default
- return site-coords per kwarg-request not per default
- ?

Thoughts?

mgrover1:
I like the idea of keeping that at the root level... we can add an attribute perhaps? But keeping at the root dataset level makes sense

kmuehlbauer:
@mgrover1 Yes, they should definitely stay at the root group. And they should not show up in sweep groups inside DataTree.

But it would be great if they could show up when single sweeps are read via `xr.open_dataset`.

I'll create a PR with a proposal.

PR #49, opened 2022-11-17
CfRadial1 reader kwargs refactor, add to_cfradial2 exporter 

kmuehlbauer:
1. CfRadial1 reader kwarg refactor:

```python
"""
    first_dim : str
        Can be ``time`` or ``auto`` first dimension. If set to ``auto``,
        first dimension will be either ``azimuth`` or ``elevation`` depending on
        type of sweep. Defaults to ``auto``.
    optional : bool
        Import optional mandatory data and metadata, defaults to ``True``.
    site_coords : bool
        Attach radar site-coordinates to Dataset, defaults to ``True``.
"""
```

```python
swp = xr.open_dataset(fname, group="sweep_0", site_coords=False, optional=False, first_dim="time", engine="cfradial1")
```

These kwargs are available for DataTree loading, as well as Dataset-loading.

2. Add `to_cfradial2` export function
    - added functionality to model to conform a sweep dataset, such that only CfRadial2 conforming data is written (needs extension and enhancement)

Todo: Add tests for the introduced functionality.

closes: #48 

~~This PR also updates pre-commit hooks:~~

~~- add nbstripout, to clean notebooks before committing~~
~~- remove the `double-quote-string-fixer`, as it contradicts with black-formatting~~ 
~~- remove the skip-string-normalization from black-formatting~~ 
~~- format with black~~ 

~~I can split this into two, if necessary. Need to make sure to do rebase-merge otherwise.~~
This part moved to #50

### DIFF
--- a/xradar/io/backends/cfradial1.py
+++ b/xradar/io/backends/cfradial1.py
@@ -94,8 +94,27 @@ def _get_required_root_dataset(ds, optional=True):
     return root
 
 
-def _get_sweep_groups(root, sweep=None, first_dim="time"):
+def _get_sweep_groups(
+    root, sweep=None, first_dim="time", optional=True, site_coords=False
+):
     """Extract Sweep Groups.
+
+    Parameters
+    ----------
+    root : xarray.Dataset
+        CfRadial1 Dataset
+
+    Keyword Arguments
+    -----------------
+    sweep : str, int, optional
+        Sweep/Group to extract, default to None.
+    first_dim : str
+        Defaults to `time` as first dimension. If set to `auto`, first dimension will
+        be either `azimuth` or `elevation` depending on type of sweep.
+    optional : bool
+        Import optional mandatory data and metadata, defaults to `True`.
+    site_coords : bool
+        Attach radar site-coordinates to Dataset, defaults to `False`.
 
     Ported from wradlib.
     """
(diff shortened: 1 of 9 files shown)


xradar/io/backends/cfradial1.py
```diff
@@ -94,9 +94,28 @@ def _get_required_root_dataset(ds, optional=True):
     return root
 
 
-def _get_sweep_groups(root, sweep=None, first_dim="time"):
+def _get_sweep_groups(
+    root, sweep=None, first_dim="time", optional=True, site_coords=False
```
mgrover1, line 98: Can we set the first_dim to auto, and the site_coords to true by default? Both of these would make dealing with this easier.

kmuehlbauer: @mgrover1 Yes, but then we would need our own dedicated cfradial1 writer to swap dims backs and drop vars/coords out.

Should I add this functionality?

mgrover1: Yeah - I think this would help!

mgrover1: It's similar to how one could use dataset.to_netcdf(), but it does not necessarily conform to the additional metadata standards as a cfradial file (dataset.to_cfradial1...)

kmuehlbauer: @mgrover1 I've added a prototype which we can build on. Need to fix/add tests, but you might already have an early look at this.

kmuehlbauer:
PR docs 

- [Sweep group kwargs](https://openradar-docs--49.org.readthedocs.build/projects/xradar/en/49/notebooks/CfRadial1_Model_Transformation.html#Sweep-groups)
- [Export/Roundtripping to_netcdf](https://openradar-docs--49.org.readthedocs.build/projects/xradar/en/49/notebooks/CfRadial1_Model_Transformation.html#Roundtrip-with-to_netcdf)
- [Export/Roundtripping to_cfradial2](https://openradar-docs--49.org.readthedocs.build/projects/xradar/en/49/notebooks/CfRadial1_Model_Transformation.html#Roundtrip-with-xradar.io.to_cfradial2)

mgrover1 (approved):

mgrover1:
@kmuehlbauer I think we can go ahead and merge this

kmuehlbauer:
Yes, we can add anything I forgot in follow up PR's.

Issue #51, opened 2022-11-19
Error when opening IRIS/Sigmet file


aladinor:
* xradar version: 0.0.8
* Python version: 3.9
* Operating System: Ubuntu 22.04.1 LTS

I am trying to read an IRIS file using `xradar.io.datatree` and `xr.open_dataset ` but I am having a dimension error.
```
import pyart
import fsspec
import xarray as xr
import xradar as xd
import boto3
import botocore
from botocore.client import Config
from datetime import datetime


def create_query(date, radar_site):
    """
    Creates a query for listing IDEAM radar files stored in AWS bucket
    :param date: date to be queried. e.g datetime(2021, 10, 3, 12). Datetime python object
    :param radar_site: radar site e.g. Guaviare
    :return: string with a
    """
    prefix = f'l2_data/{date:%Y}/{date:%m}/{date:%d}/{radar_site}/{radar_site[:3].upper()}{date:%y%m%d}'
    return prefix


def main():
    str_bucket = 's3://s3-radaresideam/'
    s3 = boto3.resource('s3',
                        config=Config(signature_version=botocore.UNSIGNED, user_agent_extra='Resource'))
    bucket = s3.Bucket('s3-radaresideam')
    query = create_query(date=datetime(2021, 10, 3, 12), radar_site='Guaviare')
    radar_files = [f'{str_bucket}{i.key}' for i in bucket.objects.filter(Prefix=f"{query}")]
    for idx, i in enumerate(radar_files[2:3]):
        file = fsspec.open_local(f'simplecache::{i}', s3={'anon': True}, filecache={'cache_storage': '.'})
       # dtree = xd.io.open_iris_datatree(file)
        ds = xr.open_dataset(file, group=1, engine="iris")

if __name__ == "__main__":
    main()

```
Then it rises this error, 
```
ValueError: conflicting sizes for dimension 'range': length 580 on 'range' and length 581 on {'azimuth': 'DBTH', 'range': 'DBTH'}
python-BaseException
```

I have dug a little bit and I found that when creating the range dimension in `iris.py` file,  line `3821` , it should create a NumPy array with a size of 581 but it instead creates a 580 NumPy array.  I tried to solve it as follows but not sure if this is the right way .

```
rng = (
            np.arange(
                range_first_bin,
                range_last_bin + 1, ## ###adding 1 here
                task["step_output_bins"],
                dtype="float32",
            )[: task["number_output_bins"]]
            / 1e2
        )
```

Cheers, 

[...]

PR #53, opened 2022-11-22
Fix Iris/Sigmet, align with CfRadial2

kmuehlbauer:
This fixes the issue with #51 and removes some inconsistencies with the common CfRadial2-based API. Also undefaulted angle reindexing and removed `a1gate` and `angle_res` from angle-coordinate.

ping @aladinor @mgrover1


xradar/io/backends/iris.py
```diff
@@ -3665,9 +3665,15 @@ def get_data_headers(self):
     def site_coords(self):
         """Return Radar Location Tuple"""
         ing_conf = self.ingest_header["ingest_configuration"]
+        lon = ing_conf["longitude_radar"]
+        lat = ing_conf["latitude_radar"]
+        lon = lon if lon <= 180 else lon - 360
+        # todo: is this correct for southern latitudes?
+        lat = lat if lat <= 180 else lon - 360
```
mgrover1, line 3672: I don't think this is right the logic for southern latitudes. Latitude will be between 0 and -90 in the southern hemisphere

mgrover1 (commented):
See the one minor comment

xradar/io/backends/iris.py
kmuehlbauer: So, I was thinking the following:

- the output from iris/sigmet file is between 0 and 360 (binary angle decoding)
- the latitude runs from equator (0deg) to the north pole (90deg) and over back to the equator again (180deg, -180deg)
- then latitude runs to the south pole (270deg, -90deg) and up to equator again (0deg)

So with this assumption this formulation would be correct although it looks very weird. I've checked with Py-Art, the same formulation is used there. But no hints in the code.

mgrover1: Ahh okay! I am not as familiar with this format... that sounds reasonable then! Sorry for the confusion!

kmuehlbauer: No worries, I'm first convinced that this is correct, if I have either proof from file or from Vaisala :grimacing:

mgrover1: Sounds good!! Yeah - holding off on merging that Py-ART PR until we get the confirmation.

mgrover1 (approved):
Approved assuming all is good with Vaisala :)

kmuehlbauer:
> Approved assuming all is good with Vaisala :)

We can still change later, if it turns out other way.

PR #57, opened 2022-11-24
MIN: align odim/gamic open-api with CfRadial1/2

kmuehlbauer:
- align odim/gamic open-api with CfRadial1/2
- fix tests
- adapt notebooks


kmuehlbauer:
If there are not complains I'll get this in early next week latest.

PR #58, opened 2022-11-24
MIN: align rainbow open-api with CfRadial1/2

kmuehlbauer:
- align rainbow open-api with CfRadial1/2
- fix tests
- adapt rainbow/cfradial1 notebooks

needs to be rebased and go in after #57 is merged


kmuehlbauer:
If there are not complains I'll get this in early next week latest.

PR #62, opened 2022-12-08
ENH: Add angle reindexing util functions, add example notebook

### LINKED ISSUE #55: Angle Reindexing
* xradar version: 0.0.8

### Description

As a legacy from wradlib we have complex code for angle reindexing in xradar's codebase.

#### High precision angle coordinates

As radar angle coordinates (`azimuth` or `elevation`) are measured constantly by different techniques of detection of the antenna pointing direction the values in the data files are mostly floating point numbers. In many cases these floating point numbers are not rounded to a certain decimal but keep the full possible range of the used dtype.

Problems of that:

- 1D angle coordinate arrays yield no equidistant vector.
- 1D angle coordinate arrays are not equivalent for different timesteps but same scan setup

#### Missing rays, duplicate or additional rays

Sometimes rays (even sectors) are missing from the dataset, sometimes there are duplicate rays. Another problem with radar data are additional rays, which I call "antenna hickup" (two rays measured with within one resolution interval).

#### What is angle reindexing?

Angle reindexing takes care of these problems by trying to determine the wanted layout from the radar metadata and the angle coordinates. With that newly created angle coordinate xarray machinery is used to reindex the radar moment data to that by nearest neighbor lookup (up to a tolerance). Missing rays will be filled with NaN.

#### Why should it be used?

For most operations this is not a real problem. It will turn into a problem, if you want to stack your xarray.Dataset radar data on a third dimension (eg. `time`, by using `open_mfdataset`). Then all coordinates need alignment to keep things simple and manageable (eg. `azimuth=[0.5, 1.5, 2.5,..., 359.5]`)

#### How should we treat it?

Currently the reindexing code relies on some internals which make things a bit hard to maintain. My suggestion would be to disentangle the reindexing code from the internals but feed the needed values as parameters. Then every reader can call this per activated `reindex_angle` kwarg.

### Discussion

I'm very open for discussion to find the most optimal way to provide that feature.

kmuehlbauer:
Closes #55 

Todo:

- [x] add tests
- [x] replace old reindexing code with new functionality

### DIFF
--- a/ci/unittests.yml
+++ b/ci/unittests.yml
@@ -20,6 +20,7 @@ dependencies:
   - pytest-doctestplus
   - pytest-sugar
   - pytest-xdist
+  - scipy
   - setuptools
   - xarray
   - xmltodict
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -32,4 +32,5 @@ notebooks/Furuno
 notebooks/Rainbow
 notebooks/Iris
 notebooks/plot-ppi
+notebooks/angle_reindexing
 ```
--- a/environment.yml
+++ b/environment.yml
@@ -20,6 +20,7 @@ dependencies:
   - xmltodict
   - dask
   - matplotlib-base
+  - scipy
   - pip:
     - git+https://github.com/openradar/open-radar-data.git
     - git+https://github.com/xarray-contrib/datatree
--- a/examples/notebooks/angle_reindexing.ipynb
+++ b/examples/notebooks/angle_reindexing.ipynb
@@ -1,0 +1,282 @@
+{
+ "cells": [
+  {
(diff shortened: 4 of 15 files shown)


kmuehlbauer:
@mgrover1 I've added a notebook with an example use case. Would be great if you or others could test that with some fitting radar data.

kmuehlbauer:
Codecov isn't getting the coverage right and does not point to the latest commit. Seems it wasn't uploaded at all. There is some discussion in the links below, how to handle that. Obviously codecov is getting rate limited every once in a while.

XRef: https://github.com/codecov/codecov-action/issues/557, https://github.com/codecov/codecov-action/issues/837

PR #81, opened 2023-02-08
FIX: ODIM_H5 backend inconsistencies

### LINKED ISSUE #80: ODIM sweep_number is off by one
* xradar version: 0.0.11
* Python version: 3.11
* Operating System: Linux

### Description

xradar tries to align to standard, so `sweep_number` starts from 0 (`sweep_0`). As `ODIM_H5` is 1-based (starting with `/dataset1`), `sweep_number` is currently wrongly decoded. The `odim`-backend based data starts therefore with `sweep_1` instead of `sweep_0`.

This does not give problems, because the iteration process over the groups are the same. Nevertheless this should be corrected.

kmuehlbauer:
closes #80 

fix several issues with the odim backend

### DIFF
--- a/tests/io/test_io.py
+++ b/tests/io/test_io.py
@@ -53,13 +53,12 @@ def test_open_cfradial1_datatree(cfradial1_file):
     azimuths = [483, 483, 482, 483, 481, 482, 482, 484, 483]
     ranges = [996, 996, 996, 996, 996, 996, 996, 996, 996]
     for grp in dtree.groups:
-        print(grp)
         # only iterate sweep groups
         if "sweep" not in grp:
             continue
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
@mgrover1 This and the other PR (#82) fix some inconsistencies I've stumbled over while working on depending `wradlib` on `xradar`.

Those who want to follow that endeavor: https://github.com/wradlib/wradlib/pull/617

I'm really happy how smooth the transition is working, let aside the little disturbances. There is not much missing in xradar and I can completely remove the compatibility layer.

kmuehlbauer:
I'll go ahead an get this simple fix in, as well as the other PR's.

PR #82, opened 2023-02-08
FIX: furuno backend inconsistencies

### LINKED ISSUE #77: _FillValue missing from FurunoBackend (azimuth/elevation)
* xradar version: 0.0.11
* Python version: 3.11
* Operating System: Linux

### Description

Writing a Furuno-based Datatree to netcdf warns with 

```python
SerializationWarning: saving variable azimuth with floating point data as an integer dtype without any _FillValue to use for NaNs
```

That's due to missing _FillValue in the backend.

kmuehlbauer:
closes #77 

fixes several issues with the furuno backend


kmuehlbauer:
I've checked with Alice from GFZ Potsdam. She provided a PPI sweep at -1deg to check the correct raw dtype of the elevetaion data. We are now sure it's `int16` instead `uint16` as written in the manual.

kmuehlbauer:
This issue was escalated to Furuno Japan. They already fixed the error in their manuals and updates will be distributed.

PR #87, opened 2023-02-15
ENH: add `spatial_ref` with `pyproj` when georeferencing, add/adapt methods/tests

kmuehlbauer:
This adds `spatial_ref` to the datasets when georeferencing. That way users can easily wrap this into cartopy and use this for georeferenced plotting.

- tests added/adapted
- added example to plot-ppi.ipynb notebook
- takes site coordinates into account (altitude) for georeferencing
- site coordinates are in default model Dataset (Locarno Monti, MeteoSvizzera coords, birthplace of xradar)

### DIFF
--- a/xradar/georeference/transforms.py
+++ b/xradar/georeference/transforms.py
@@ -1,20 +1,43 @@
 #!/usr/bin/env python
-# Copyright (c) 2022, openradar developers.
+# Copyright (c) 2022-2023, openradar developers.
 # Distributed under the MIT License. See LICENSE for more info.
+
+"""
+Transformations
+^^^^^^^^^^^^^^^
+
+.. autosummary::
+   :nosignatures:
+   :toctree: generated/
+
+   {}
+"""
+
+__all__ = ["antenna_to_cartesian", "get_x_y_z", "get_x_y_z_tree"]
+
+__doc__ = __doc__.format("\n   ".join(__all__))
+
 
 import numpy as np
 import xarray as xr
 
+from .projection import add_crs, get_crs, get_earth_radius
+
 
 def antenna_to_cartesian(
-    ranges, azimuths, elevations, earth_radius=6371000, effective_radius_fraction=None
+    ranges,
+    azimuths,
+    elevations,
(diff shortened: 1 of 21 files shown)


mgrover1 (commented):
One small suggestion here - can we call it `add_crs` instead of `write_crs`?  Also, by default, can we call this when "georeference" the data?

xradar/georeference/transforms.py
```diff
@@ -95,6 +98,10 @@ def get_x_y_z(ds, earth_radius=6371000, effective_radius_fraction=None):
     .. [1] Doviak and Zrnic, Doppler Radar and Weather Observations, Second
         Edition, 1993, p. 21.
     """
+    if earth_radius is None:
```
kmuehlbauer, line 101: @mgrover1 Actually it is written in any case, when georeferenced. So the user doesn't have to do anything, beside georeferencing.

kmuehlbauer:
@mgrover1 I've refactored the naming per your suggestion.

kmuehlbauer:
@mgrover1 I'm ironing out the CI/doc issues. We can let this sit another while. No rush for now to get this in.

kmuehlbauer:
This is the direct link to the docs with a plotting example:

https://openradar-docs--87.org.readthedocs.build/projects/xradar/en/87/notebooks/plot-ppi.html#Plot-PPI-on-map-with-cartopy

kmuehlbauer:
@mgrover1 Heads-up, I've found some problems with the docs which are slightly related to this PR and I'll add that here. I'll also add some more code to model.py for the sitecoords.

kmuehlbauer:
I've also found an issue that get_x_y_z didn't take the radar site altitude into account, I've fixed this too.

mgrover1 (approved):
Great!

kmuehlbauer:
@mgrover1 I'll merge tomorrow, if no-one beats me to it.

kmuehlbauer:
I've rearranged/squashed the commits into two for rebase-merge. Getting this when CI is green.

PR #95, opened 2023-03-01
ENH: make copy before overwriting, add compression in to_odim

kmuehlbauer:
- [ ] Closes #xxxx
- [x] Tests added

This fixes an issue when writing `to_odim` if the underlying data array's write flag wasn't set. It also adds the capability to add `compression`/`compression_opts` kwargs in the call to `to_odim`.

The latter makes it very easy and convenient to use the [zoo of compressors](https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins) available to `hdf5`. Please refer to http://www.silx.org/doc/hdf5plugin/latest/index.html for usage examples.


mgrover1 (approved):
Great! Looks good to me.

PR #96, opened 2023-03-03
ENH add mandatory source parameter to odim writer

egouden:
Here is a small improvement

### DIFF
--- a/xradar/io/export/odim.py
+++ b/xradar/io/export/odim.py
@@ -134,7 +134,7 @@ def _write_odim_dataspace(source, destination, compression, compression_opts):
             ds.attrs.create("IMAGE_VERSION", version, dtype=H5T_C_S1_VER)
 
 
-def to_odim(dtree, filename, compression="gzip", compression_opts=6):
+def to_odim(dtree, filename, source=None, compression="gzip", compression_opts=6):
     """Save DataTree to ODIM_H5/V2_2 compliant file.
 
     Parameters
@@ -145,12 +145,22 @@ def to_odim(dtree, filename, compression="gzip", compression_opts=6):
 
     Keyword Arguments
     -----------------
+    source : str
+        mandatory radar identifier (see ODIM documentation)
     compression : str
         Compression filter name, defaults to "gzip".
     compression_opts : compression strategy
         options as needed by above filter, defaults to 6
-
     """
+    nod = "NOD" in source
+    wmo = "WMO" in source
+    rad = "RAD" in source
+    if not nod and not rad and not wmo:
+        raise ValueError(
+            "Please provide the source parameter with at least one"
+            "of the mandatory radar identifier (NOD, RAD, WMO)"
+        )
+
     root = dtree["/"]
 
     h5 = h5py.File(filename, "w")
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
Thanks @egouden for adding this missing mandatory part.

I'm linking the [ODIM_H5 2.2](https://www.eumetnet.eu/wp-content/uploads/2019/05/OPERA-ODIM_H5-v2.2.pdf) document for reference.

Below is table 3 of the linked document, showing some description of the possible content.

Table_3

We might want to check if at least one of the listed descriptors (eg. `WMO`) is given in the `source`-kwarg. Just to make sure we really write standard conform metadata. 

For the error in CI (lint and style): We are using [black](https://black.readthedocs.io/en/stable/index.html) and [ruff](https://beta.ruff.rs/docs/) for linting. To automate the process a bit we also added a [.pre-commit-config.yaml](https://github.com/openradar/xradar/blob/main/.pre-commit-config.yaml) file, which can be installed in your local xradar git repo (refer to [pre-commit docs](https://pre-commit.com/)). If the linter is happy unit-tests are run.

xradar/io/export/odim.py
```diff
@@ -145,12 +145,17 @@ def to_odim(dtree, filename, compression="gzip", compression_opts=6):
...
+    source : str
+        radar identifier (see ODIM documentation)
     compression : str
         Compression filter name, defaults to "gzip".
     compression_opts : compression strategy
         options as needed by above filter, defaults to 6
-
     """
+
+    if source is None:
```
kmuehlbauer, line 156: Here we might add that check, if at least one of those descriptors is given with `source`.

kmuehlbauer (changes requested):
This is really good to see this fixed, after all these years going unnoticed in wradlib. Thanks @egouden and welcome as contributor.

I've added two comments where we could slightly improve.

egouden:
Thank you @kmuehlbauer for the quick review.

I added an explicit check for mandatory radar identifiers. I would not add more complex tests at this stage.

I am not testing reading back the written file because I see no direct equivalent in the model based on FM301.

kmuehlbauer:
Thanks @egouden!

> I added an explicit check for mandatory radar identifiers. I would not add more complex tests at this stage.

Yes, we can iterate on that when needed later.

> I am not testing reading back the written file because I see no direct equivalent in the model based on FM301.

You're right, there is no direct equivalent in the current standard. That's something we might solve by adding to xarray's `encoding` attribute. This needs some thought, though. Would you mind opening an issue for that, so we do not forget about it?

The PullRequest is looking good to me. Please add an entry in https://github.com/openradar/xradar/blob/main/docs/history.md. I will let this sit another day or two, so that others can have a look too, before merge.

kmuehlbauer:
> I would like to add a test for the exception. Do you have any recommendations?

You can try to call `to_odim` with broken source-kwargs.

Then you can use [pytest-raises](https://docs.pytest.org/en/4.6.x/reference.html#pytest-raises) to see if the Exception was raised correctly.

kmuehlbauer:
@egouden Let me know if you are fine with it now. I'll have @mgrover1 have another look before merge.

kmuehlbauer (approved):

mgrover1 (approved):
Looks great!

kmuehlbauer:
Thanks @egouden, welcome on board xradar.

PR #97, opened 2023-03-08
ENH: switch to add recommended metadata

egouden:
Now by default the function to_odim writes only mandatory metadata. There is a new option to add recommended metadata. See ODIM documentation for more information.

Note that only a few recommended metadata can be written at the moment.

### DIFF
--- a/xradar/io/export/odim.py
+++ b/xradar/io/export/odim.py
@@ -134,7 +134,14 @@ def _write_odim_dataspace(source, destination, compression, compression_opts):
             ds.attrs.create("IMAGE_VERSION", version, dtype=H5T_C_S1_VER)
 
 
-def to_odim(dtree, filename, source=None, compression="gzip", compression_opts=6):
+def to_odim(
+    dtree,
+    filename,
+    source=None,
+    optional_how=False,
+    compression="gzip",
+    compression_opts=6,
+):
     """Save DataTree to ODIM_H5/V2_2 compliant file.
 
     Parameters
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
Thanks @egouden for pushing this forward.

Would you think it useful to have a kwarg `metadata` instead, which can have:

- `mandatory`
- `recommended`
- `all`

egouden:
The problem is a bit more complex. In ODIM 2.2 you have the concept of  "recommended" and "prioritized" how attributes. In ODIM 2.3, you have the corresponding "recommended" and "highly desirable". In ODIM 2.4, it becomes"optional" and "mandatory".

I was considering the option of a keyword argument too. However, I see only two options: "mandatory" and "all". Therefore I think that having a switch is more clear for both users and developers. 

I just changed the name of the switch to a more explicit "optional_how", also for forward compatibility.

xradar/io/export/odim.py
```diff
@@ -147,6 +154,8 @@ def to_odim(dtree, filename, source=None, compression="gzip", compression_opts=6
     -----------------
     source : str
         mandatory radar identifier (see ODIM documentation)
+    optional_how : boolean
+        true to include optional how attributes
```
kmuehlbauer, line 158: Please add: "defaults to False" here.

kmuehlbauer:
Thanks @egouden for the details.

We should elaborate on better version handling in `to_odim` later. I'll open an issue for that.

egouden:
The requested changes have been implemented.

kmuehlbauer:
@egouden Would you mind adding to history.md? We can merge right after that. Thanks!

egouden:
The history has been updated.

kmuehlbauer:
Thanks @egouden! 

I'd consider those enhancement/fixes of yours worth a new xradar release. WDYT?

PR #110, opened 2023-05-17
FIX: Correct DB_DBTE8/DB_DBZE8 and DB_DBTE16/DB_DBZE16 decoding

kmuehlbauer:

Update: 

A bit more context here, we've just stumbled over these types in the wild. I've had a look into the docs/code and they are referenced as:

```
 DB_DBTE8         = 71,  /* Total Power Enhanced (via H+V or HV) (1 byte) */
 DB_DBTE16        = 72,  /* Total Power Enhanced (via H+V or HV) (2 byte) */
 DB_DBZE8         = 73,  /* Clutter Corrected Reflectivity Enhanced (1 byte) */
 DB_DBZE16        = 74,  /* Clutter Corrected Reflectivity Enhanced (2 byte) */
```

I've not added this to the moments mapping (attributes). So for now no attributes will be added. We'll gather more information before adding these attributes.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -2,6 +2,7 @@
 
 ## Development Version
 * ENH: Add new examples using radar data on AWS s3 bucket ({pull}`102`) by [@aladinor](https://github.com/aladinor)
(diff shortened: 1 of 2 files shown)


mgrover1:
@kmuehlbauer let me know when this ready for review

PR #114, opened 2023-05-25
FIX ipol_time not working if last azimuth is first

egouden:
The function util.ipol_time is not working if the last azimuth is the first timestamp. This causes a zero-length array. You might want to see the change itself for more insight.


kmuehlbauer:
Thanks @egouden. Can you add a test for this?

egouden:
It is unclear to me what you would like to test.

kmuehlbauer:
A test which fails with current code and works with your fix

kmuehlbauer:
Superseded by #117. Thanks @egouden for getting this into traction.

PR #117, opened 2023-06-13
FIX: time_interpolation

kmuehlbauer:
- [x] Tests added

As reported by @egouden in #114 there are issues with time interpolation in some corner cases. While adding a test for this I've found there are more things involved which need changing. This PR addresses (hopefully) all of this.

- only interpolate of NaT present
- handle NaT correctly (-9223372036854775808)
- handle a1gate properly (roll data to first measured ray)
- derive interpolation angles using diff
- handle wrap around angles for PPI
- fix spurious digits introduced by floating point interpolation

Supersedes #114


mgrover1 (approved):
This looks great @kmuehlbauer - thanks for catching this/dealing with this!

kmuehlbauer:
OK, let's get this in then.

PR #125, opened 2023-09-14
FIX: always read nodata and undetect attributes from ODIM file

egouden:
The current behavior is to not read ODIM _undetect_ and _nodata_ attributes if the values are not transformed (i.e. the attribute _gain_ is 1 and the attribute _offset_ is 0). This can lead to a loss of information. I think these attributes should always be read.


kmuehlbauer:
@egouden Sorry for the delay here. Could you please fix the linter error. Otherwise this looks OK to me.

mgrover1 (approved):
This looks great!!

kmuehlbauer (approved):

kmuehlbauer:
Thanks @egouden!

PR #127, opened 2023-09-25
use `crs_wkt` instead of deprecated `spatial_ref`

kmuehlbauer:
This forces writing of `crs_wkt` instead of deprecated `spatial_ref`. On reading `crs_wkt` takes precedence over `spatial_ref` (for backwards compatibility).

- [x] Tests fixed


mgrover1 (approved):
this makes a lot of sense - thank you @kmuehlbauer

PR #128, opened 2023-09-25
use `cmweather` colormaps in xradar

kmuehlbauer:
Would be nice, if we advertise the usage of colorblind friendly colormaps.


mgrover1 (approved):
This looks great @kmuehlbauer ! Thank you!!

PR #132, opened 2023-09-29
FIX: _FillValue + History in the cfradial1 exporter

syedhamidali:


docs/history.md
```diff
@@ -1,7 +1,7 @@
 # History
 
 ## 0.4.0 (2023-09-27)
-
+* FIX: Add _FillValue to CfRadial1 exporter, and minor error in CfRadial1_Export.ipynb notebook by [@syedhamidali](https://github.com/syedhamidali)
```
kmuehlbauer, line 4: Could you please ad a new section

```python
## Development version (unreleased)
```

and move your fix there?

xradar/io/export/cfradial1.py
```diff
@@ -304,6 +306,21 @@ def to_cfradial1(dtree=None, filename=None, calibs=True):
 
     dataset.attrs = dtree.attrs
 
+    # Find variables with >= 2 dimensions
+    multidim_variables = [
```
kmuehlbauer, line 310: Do we need to check if we have `azimuth`/`range`, `elevation`/`range` or `time`/`range` as dimensions?

xradar/io/export/cfradial1.py
```diff
@@ -304,6 +306,21 @@ def to_cfradial1(dtree=None, filename=None, calibs=True):
...
+        var for var in dataset.data_vars if len(dataset[var].dims) >= 2
+    ]
+
+    # Set _FillValue attribute to -32768 for variables with >= 2 dimensions
+    fill_value = -32768
```
kmuehlbauer, line 315: Is this _FillValue documented somewhere? Is it in the CfRadial1 standard?

kmuehlbauer (changes requested):
@syedhamidali Thanks for continuing to shape the CfRadial1 exporter. I've added some remarks and questions.

xradar/io/export/cfradial1.py
syedhamidali: Yeah, so I designed it according to the file DATASETS.fetch("cfrad.20080604_002217_000_SPOL_v36_SUR.nc"). And, I have also noticed that number -32768 in other CfRadial data before this. I think this is the most common fillvalue used in the CfRadial1 data. Did I answer your question?

xradar/io/export/cfradial1.py
syedhamidali: I don't think we need to confirm these dims explicitly, although it will be better if we do that, but again we are doing that in the _variable_mapper function. Generally in the CfRadial1 data, there will be data variables with the same shape based on the number of rays (nrays) and gates (ngates). It doesn't matter whether these are base moments, polarimetric, or even reterivals, they should have this fill value. But, if you would like to see it that way, then i can add that condition easily.

kmuehlbauer: Great, let's keep this as is and react if needed.

xradar/io/export/cfradial1.py
kmuehlbauer: I've quickly swept the CfRadial docs and there is no default `_FillValue` mentioned.

I'd suggest to use the netcdf default `_FillValue` in these cases and probably make this parametrizable.

@mgrover1 Do you have a bit more insight here. Maybe ping one of the CfRadial experts?

kmuehlbauer (approved):
@syedhamidali This is OK with me. I've one issue with the `_FillValue` but will rely on the expertise of @mgrover1 here. Thanks!

xradar/io/export/cfradial1.py
syedhamidali: Thanks for this info! Also, what is the default `_FillValue` for the netcdf?

syedhamidali: This is where I took it from!
Screenshot 2023-10-02 at 12 14 27 PM

kmuehlbauer: You can get it from `netCDF4.default_fillvals` (netcdf4-python). Or you can grab it from `netcdf.h` in the netcdf-c sources ;-)

mgrover1:
There is no "official" fill value - that is left up to the data providers!

xradar/io/export/cfradial1.py
syedhamidali: > You can get it from `netCDF4.default_fillvals` (netcdf4-python). Or you can grab it from `netcdf.h` in the netcdf-c sources ;-)

Great, thank you! I'll take a look right away.

syedhamidali: If you perform a GitHub search for queries like `pyart "-32768"` or `lrose "-32768"`, you'll notice that in the code sections of various radar tools such as Pyart, Lrose Radx, Singledop, Multidop, Meteoswiss, and more, the value "-32768" is commonly utilized. However, as @mgrover1 said, we can use `netcdf default` as well.

kmuehlbauer:
> There is no "official" fill value - that is left up to the data providers!

Then make it parametrizable? And use netcdf default if not specified?

kmuehlbauer:
@syedhamidali OK, maybe we can skip that _FillValue issue altogether.

If _FillValue is not given, netcdf4 will take the default value anyway. If _FillValue is set, it will take this _FillValue. So maybe we should just document?

syedhamidali:
> @syedhamidali OK, maybe we can skip that _FillValue issue altogether.
> 
> If _FillValue is not given, netcdf4 will take the default value anyway. If _FillValue is set, it will take this _FillValue. So maybe we should just document?

Actually, I tried to manually check the same using pyart prior to this PR, and I noticed the fill values don't get filled in. That's one of the reason for this PR.

syedhamidali:
@kmuehlbauer , I have dropped the `__FillValue`, as it is taken care of by default netcdf methods maybe. I think we are good to go with this PR. The changes include: 
-  Minor error fixed in the Notebook
-  Added xradar version and cfradial1 version to the dataset attributes

mgrover1 (approved):
This looks great to me!

kmuehlbauer (approved):
LGTM! Thanks @syedhamidali !

Issue #133, opened 2023-10-04
DataTree to Dataset transformation method


syedhamidali:
### Proposal to Implement Data Transformation Method from **cfradial2** to **cfradial1** and vice versa

Dear Xradar Team,

I would like to propose the introduction of a data transformation method that facilitates smooth conversion from the current **cfradial2** format to the legacy **cfradial1** format within the Xradar library. This addition would greatly enhance user experience and improve compatibility with existing code. Specifically, it would enable users to perform operations on the entire volume, i.e., across all radar sweeps. However, it's essential to ensure that these operations are reversible, allowing for seamless transitions between **datatree to dataset** and **dataset to datatree** formats.

The proposed method would be accessible through either of the following options:

- Using a dedicated method, e.g., `ds = xd.to_dataset(dtree)` where `dtree` is a **cfradial2/datatree.DataTree** object, and `dtree = xd.to_datatree(ds)`, where ds is **xarray.Dataset**
Or,
- Implementing a custom method, `ds = dtree.to_dataset()`, if `dtree` is an Xradar object, and similarly `ds.to_datatree` if ds is an cfradial1 dataset. However, these approaches may pose some challenges as these are not directly derived from the class, but would be a great feature.

To address this, I suggest creating a new module named `cfradial1_model.py` within the root directory. This module would incorporate the code currently residing in `xradar/io/export/cfradial1.py` and provide a method to return the dataset exclusively. This approach would ensure a clean separation of concerns.

Additionally, for the export `to_cfradial1` functionality, we could import the necessary function from `cfradial1_model.py` and extend it with the option to save the dataset to a NetCDF file, like so: `ds.to_netcdf(file)`. I can easily take care of datatree to dataset transformation, but would need help with the datatree to dataset transformation.

By adopting this proposal, I aim to simplify the process of transitioning between **cfradial2** to **cfradial1** formats, and vice versa, within the Xradar library, without affecting the current method to export the cfradial1 data. 

Your feedback and suggestions are highly appreciated, and looking forward to your thoughts on this proposal.

Best,
Hamid

mgrover1:
I agree this functionality would be helpful!

kmuehlbauer:
I totally support this. I have one suggestion so far. The function naming is too general IMHO. What about ``.to_cf1_dataset()` and `to_cf2_datatree()` or even shorter `.to_cf1()`, `to_cf2()`? 

We would need to be careful in the creation process wrt data copies/views. That might need some thinking on the use cases.

[...]

PR #136, opened 2023-10-23
Convert DataTree to Dataset and vice versa.

syedhamidali:
- [ ] Closes #xxxx
- [ ] Tests added

Building upon issue #133 
I kindly request a thorough review of this pull request. It will alter the structure of DataTree (`-> datatree.Cfradial2`) and Dataset (`-> xarray.Cfradial1`), with some code duplication from `backends` to the `transform` modules. Your assistance in finalizing the better structure would be greatly appreciated.

Moreover, by structuring these functions within a class, it paves the way for future enhancements beyond conversion (e.g., `dtree.to_cfradial1_dataset()` or `ds.to_cf2()`) and exporting (`ds.to_netcdf()`, maintaining the `cfradial1` data standard), including features like `dtree.plot_ppi()`. While this may be a potential future development, these changes currently encompass our modifications.


kmuehlbauer:
@syedhamidali First of all, thank you for taking on that difficult matter.

I'm a bit hesitant when it comes to subclassing. Xarray explicitely encourages downstream users to use [composition over inheritance](https://docs.xarray.dev/en/stable/internals/extending-xarray.html#composition-over-inheritance). And this aligns with my experience from the early wradlib xarray experiments.

Another aspect is that we might not be able to keep the requirements for being a CfRadial1/CfRadial2 class instance throughout it's lifetime. During processing relevant coordinates/variables/attributes might just get dropped or otherwise tampered with and subsequent calls to the conversion functions will fail. We would have to make sure that the class instance still contains what it should. I'm not sure we want to take on that burden to check every change in the class instance if we still have a valid CfRadial1/CfRadial2 object.

Instead I would favour keeping `xarray.Dataset` and `datatree.DataTree` and just implementing the conversion functions, which then could be accessed by accessors.

syedhamidali:
@kmuehlbauer Yeah, I completely agree with you, I am also hesitant to use custom subclassing, and I appreciate you sharing the helpful link. I've gained some initial insights into "extending xarray using accessors" from it. I'll explore this further, but if I run into any roadblocks, I may reach out to you or @mgrover1 for assistance. Thank you!

syedhamidali:
Closing this PR without merging, see #224

PR #138, opened 2023-10-25
FIX: prevent integer overflow when calculating azimuth in FURUNO scn files

### LINKED ISSUE #137: integer overflow when calculating azimuth in FURUNO scn files
* xradar version: 0.4.0
* Python version: 3.11
* Operating System: Ubuntu 22.0.4

### Description

Problem with calculation of the azimuth values in a specific file at these lines:
https://github.com/openradar/xradar/blob/6880b0cbfaff92a084e07bfefcc0f63bdc826535/xradar/io/backends/furuno.py#L365-L366

[1010_radar.scn[22020].gz](https://github.com/openradar/xradar/files/13121757/1010_radar.scn.22020.gz)

### What I Did
If you look this post, you can see what i did

https://openradar.discourse.group/t/problem-with-snc-file/335/3

kmuehlbauer:
- [x] Closes #137 

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -3,6 +3,7 @@
 ## Development version (unreleased)
 
 * FIX: Add history to cfradial1 output, and fix minor error in CfRadial1_Export.ipynb notebook({pull}`132`) by [@syedhamidali](https://github.com/syedhamidali)
+* FIX: prevent integer overflow when calculating azimuth in FURUNO scn files ({issue}`137`) by [@giacant](https://github.com/giacant) , ({pull}`138`) by [@kmuehlbauer](https://github.com/kmuehlbauer)
 
 ## 0.4.0 (2023-09-27)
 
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
We can't really test this without having a file which contains that issue. Or we would need to think to refactor the code into functions we can test. Until now we just might have to live with the current state of affairs.

kmuehlbauer:
@mgrover1 I'll go ahead and merge tomorrow. Will also do a new release then.

giacant:
> We can't really test this without having a file which contains that issue. Or we would need to think to refactor the code into functions we can test. Until now we just might have to live with the current state of affairs.

I can give you more files with this problem if you need

kmuehlbauer:
https://docs.openradarscience.org/projects/xradar/en/stable/installation.html

Section: From sources. Just clone from my fork and check out this very branch.

kmuehlbauer:
Yep, just create a new environment for that test. You can throw it away after that.

kmuehlbauer:
@giacant That version is the latest version on main.

Try that, it should directly install this branch.

`python -m pip install git+https://github.com/kmuehlbauer/xradar.git@fix-137`

giacant:
OK, now it works!

PR #139, opened 2023-10-25
FIX: make coordinates consistent

kmuehlbauer:
Remove unneeded coordinates.

- [x] Changes are documented in `history.md`

PR #140, opened 2023-10-25
FIX: fix sphinx-favicon, fix sphinx-extlinks to get readthedocs running

kmuehlbauer:


kmuehlbauer:
The current unit-test errors will be fixed in #139.

PR #142, opened 2023-10-31
FIX: explicitely check for "False" in get_crs()

kmuehlbauer:

Currently this will always return AEQD crs instead of the one in crs_wkt since the current check just checks for `True`. It fails since crs_wkt is a variable with value 0.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -1,7 +1,8 @@
 # History
 
 ## Development Version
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@mgrover1 This one bit me over at wradlib today. Somehow my current check always uses the AEQD :grimacing: I'm going to merge this on Thursday latest and will do another bug-fix release (0.4.2) if no-one beats me to it.

PR #143, opened 2023-10-31
FIX: Fix handling of sweep_mode attribiutes

mgrover1:
We currently do not decode the sweep_mode attributes properly with the cfradial reader, and the common decoder used in the readers does not fully handle xr.DataArray data structures. This adds a fix, along with a test to ensure we are handling this properly

- [ ] Tests added


kmuehlbauer (approved):
LGTM!

PR #147, opened 2024-01-05
ENH: Add a reader for nexrad level2 files

### LINKED ISSUE #40: Add a NEXRAD reader
### Description

Add a backend for NEXRAD level 2 and level 3 files

### What We Want to Do

NEXRAD data is heavily used within the radar community, and it would be helpful to have a reader for this data in xradar.

Py-ART has readers for this, including the
- [NEXRAD Level 2 Reader](https://github.com/ARM-DOE/pyart/blob/main/pyart/io/nexrad_archive.py)
- [NEXRAD Level 3 Reader](https://github.com/ARM-DOE/pyart/blob/main/pyart/io/nexrad_level3.py)

@jthielen mentioned he would be interested in helping with this (see #27)... others can help too!

mgrover1:
This is a first cut at the nexrad file reader... starting with level2 data. Still a work in progress for now, but I figured I would share what I have so far. I may have more questions about backends @kmuehlbauer and how to deal with loading in the py-art like dictionaries.

- [x] Closes #40 
- [x] Tests added


mgrover1:
@kmuehlbauer - I am still struggling to get this up and running... do you think there would be some utility to have a generic `RadarDataStore` object, similar to the `pyart.base.Radar` object that could port dictionaries --> xarray backends?

For example, taking the following as input:

RadarDataStore(
        time,
        _range,
        fields,
        metadata,
        scan_type,
        latitude,
        longitude,
        altitude,
        sweep_number,
        sweep_mode,
        fixed_angle,
        sweep_start_ray_index,
        sweep_end_ray_index,
        azimuth,
        elevation,
        instrument_parameters=instrument_parameters,
    )

It would return an xarray data structure? I am having trouble decoupling the entry point commonalities from the individual file parsing structures in the current backends...

kmuehlbauer:
@mgrover1 I've had not yet time to check this PR out. Hopefully I can free up some time next week.

If such object makes code readability and maintenance easier, why not. How would that be integrated into xarray-backend machinery?

mgrover1:
My thought right now is that it would be structured as:

```python
NexradLevel2File --> RadarDataStore --> NexradBackendEntrypoint
```

The benefit here would be that the RadarDataStore would be the primary object that we would fit the coordinates + fields into... then we can pass that into the backend entrypoint. I can prototype this in this PR, and ping you when it is ready for feedback?

mgrover1:
@kmuehlbauer - I ended up going with a function instead of a class... it takes in the same things the radar object did in Py-ART, and returns an xarray.Dataset, with a group argument that can be used to specify which sweep to use... this way, it can be used directly with the backend entrypoint API... the user can then add additional bits to their dataset before returning that to the user. Open to thoughts here!

mgrover1:
@kmuehlbauer - I am stumped on what I am doing wrong here for it not to recognize the cython submodule I added.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1458 @@
...
+    interpolate = _find_scans_to_interp(scan_info, first_gate, gate_spacing)
+
+    # Deal with time
+    time_start, _time = nfile.get_times(scans)
+    time = get_time_attrs(time_start)
+    time["data"] = _time
+
+    # range
+    _range = get_range_attrs()
+    first_gate, gate_spacing, last_gate = _find_range_params(scan_info)
```
kmuehlbauer, line 1084: duplicate of line 1072?

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1458 @@
...
+        else:
+            i = i
+        fixed_agl.append(i)
+    fixed_angles = np.array(fixed_agl, dtype="float32")
+    fixed_angle["data"] = fixed_angles
+
+    # fields
+    max_ngates = len(_range["data"])
+    available_moments = {m for scan in scan_info for m in scan["moments"]}
+    interpolate = _find_scans_to_interp(
```
kmuehlbauer, line 1193: Looks like this is already done above at line 1075?

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1458 @@
...
+            dic = LazyLoadDict(dic)
+            data_call = _NEXRADLevel2StagedField(nfile, moment, max_ngates, scans)
+            dic.set_lazy("data", data_call)
+        else:
+            mdata = nfile.get_data(moment, max_ngates, scans=scans)
+            if moment in interpolate:
+                interp_scans = interpolate[moment]
+                warnings.warn(
+                    "Gate spacing is not constant, interpolating data in "
+                    + f"scans {interp_scans} for moment {moment}.",
```
kmuehlbauer, line 1213: It seems, that we do not have to do interpolation here. AFAICT this is only needed for CfRadial1 data  (like good old Py-ART data model). We should be safe to keep the sweep resolution as is.

kmuehlbauer (commented):
@mgrover1 Thanks for moving this forward. I've added a couple of comments.

I'm still not convinced we need the cythonized interpolation at all. It looks like it is only needed to conform the single sweeps onto a common range resolution. This won't be needed for our sweep based data model. Or am I missing something?

So in case my assumption is correct, I'd suggest to shape the code to keep the original sweep resolution and remove all Cython related things from this PR.

xradar/io/backends/nexrad_level2.py
mgrover1: I removed the interpolation in the latest commit @kmuehlbauer :) thanks for the suggestion here.

xradar/io/backends/nexrad_level2.py
mgrover1: See the latest commit - that should fix the duplication

xradar/io/backends/nexrad_level2.py
mgrover1: Yep - should be fixed

xradar/io/backends/__init__.py
```diff
@@ -24,5 +25,7 @@
 from .iris import *  # noqa
 from .odim import *  # noqa
 from .rainbow import *  # noqa
-
-__all__ = [s for s in dir() if not s.startswith("_")]
```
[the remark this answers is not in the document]
mgrover1: Not sure... it is back in. thanks!!

xradar/io/backends/nexrad_common.py
```diff
@@ -0,0 +1,248 @@
+"""
+Data and functions common to all types of NEXRAD files.
+
+"""
+
+# The functions in this module are intended to be used in other
+# nexrad related modules. The functions are not and should not be
+# exported into the pyart.io namespace.
```
kmuehlbauer, line 8: Maybe rephrasing the provenience here, so occasional users see that this is xradar with functionality forked/vendored from Py-ART.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+        # return raw data if requested
+        if raw_data:
+            return data
+
+        # mask, scan and offset, assume that the offset and scale
+        # are the same in all scans/gates
+        for scan in scans:  # find a scan which contains the moment
+            msg_num = self.scan_msgs[scan][0]
+            msg = self.radial_records[msg_num]
+            if moment in msg.keys():
```
kmuehlbauer, line 539: Would it make sense, to leave the CF encoding/decoding to xarray? We would need to add according attributes to the variables, though. I think we can take care of this in a follow up PR.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+        max_ngates : int
+            Maximum number of gates (bins) in any ray.
+            requested.
+        raw_data : bool
+            True to return the raw data, False to perform masking as well as
+            applying the appropiate scale and offset to the data.  When
+            raw_data is True values of 1 in the data likely indicate that
+            the gate was not present in the sweep, in some cases in will
+            indicate range folded data.
+        scans : list or None.
```
kmuehlbauer, line 502: This seems like we should be double careful here to not read two sweeps with different gate-spacing here. Do we have a test file which actually has different gate-spacing?

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+        data = np.ones((nrays, max_ngates), ">B")
+        for i, msg_num in enumerate(msg_nums):
+            msg = self.radial_records[msg_num]
+            if moment not in msg.keys():
+                continue
+            if not set_datatype:
+                data = data.astype(">" + _bits_to_code(msg, moment))
+                set_datatype = True
+
+            ngates = min(msg[moment]["ngates"], max_ngates, len(msg[moment]["data"]))
```
kmuehlbauer, line 528: This is connected with different gate-spacing, double checking this works as intended with a test would be great.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+        # are the same in all scans/gates
+        for scan in scans:  # find a scan which contains the moment
+            msg_num = self.scan_msgs[scan][0]
+            msg = self.radial_records[msg_num]
+            if moment in msg.keys():
+                offset = np.float32(msg[moment]["offset"])
+                scale = np.float32(msg[moment]["scale"])
+                mask = data <= 1
+                scaled_data = (data - offset) / scale
+                return np.ma.array(scaled_data, mask=mask)
```
kmuehlbauer, line 544: We might also get rid of the `mask` and masked array here, if we correctly specify missing values and/or _FillValues as attributes. Can be done as follow-up PR.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+    scans=None,
+    storage_options={"anon": True},
+    first_dimension=None,
+    group=None,
+    **kwargs,
+):
+    # Load the data file in using NEXRADLevel2File Class
+    nfile = NEXRADLevel2File(
+        prepare_for_read(filename, storage_options=storage_options)
+    )
```
kmuehlbauer, line 1019: Can we please make this depending on `storage_options`. 

```suggestion
    storage_options=None,
    first_dimension=None,
    group=None,
    **kwargs,
):
    # Load the data file in using NEXRADLevel2File Class
    if storage_options is not None:
        filename = prepare_for_read(filename, storage_options=storage_options)
    nfile = NEXRADLevel2File(
        filename
    )
```

Maybe this is a bit more involved, as `storage_options` would have to be traversed to the backend reader.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+
+    # Deal with time
+    time_start, _time = nfile.get_times(scans)
+    time = get_time_attrs(time_start)
+    time["data"] = _time
+
+    # range
+    _range = get_range_attrs()
+    first_gate, gate_spacing, last_gate = _find_range_params(scan_info)
+    _range["data"] = np.arange(first_gate, last_gate, gate_spacing, "float32")
```
kmuehlbauer, line 1033: Oh my, this is really hard to move from the CfRadial1 to CfRadial2 data model. AFAICS `_range`-dict is used for all sweeps (assuming range interpolated to a common grid). This would need to be done on a per sweep basis.

This would need another round of refactoring here.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+    # fields
+    max_ngates = len(_range["data"])
+
+    fields = {}
+    for moment in available_moments:
+        dic = get_moment_attrs(nexrad_mapping[moment])
+        dic["_FillValue"] = -9999
+        if delay_field_loading:
+            dic = LazyLoadDict(dic)
+            data_call = _NEXRADLevel2StagedField(nfile, moment, max_ngates, scans)
```
kmuehlbauer, line 1147: Here is the pain point, this will extract the moments of different scans (which might be on different range resolutions.

xradar/io/backends/nexrad_level2.py
```diff
@@ -0,0 +1,1381 @@
...
+        sweep_end_ray_index,
+        azimuth,
+        elevation,
+        instrument_parameters,
+        first_dimension=first_dimension,
+        group=group,
+    )
+
+
+def create_dataset_from_fields(
```
kmuehlbauer, line 1176: This whole function assumes that all data of all sweeps is in a common range grid. This would need refactor to work on a per sweep basis.

kmuehlbauer (changes requested):
@mgrover1 We knew that it would not be an easy task. I've added another couple of suggestions and ideas. 

It looks like we need to tackle the different gate spacing stuff at a lower level.

mgrover1:
> @mgrover1 We knew that it would not be an easy task. I've added another couple of suggestions and ideas.
> 
> It looks like we need to tackle the different gate spacing stuff at a lower level.

Thanks for the suggestions - I am busy with the AMS conference today, but will follow up later this week. I appreciate the feedback! Agreed - it is not easy, but will be worth it :)

mgrover1:
Closing this since @kmuehlbauer refactored + submitted with #158

PR #158, opened 2024-03-06
NEXRAD Level2 structured reader 

### LINKED ISSUE #160: ENH: Add a NEXRAD Level2 Reader
### Description

Add a reader for NEXRAD Level2 data, based on the one currently in the Python ARM Radar Toolkit (Py-ART)

kmuehlbauer:
- [ ] Closes #160 
- [x] Tests added

This is a rewrite of #147 using parts of the structured reader of the iris/sigmet reader. This also enhances the ability of reading much more metadata from the nexrad level 2 archive.

Note: this only implements MSG_31 as of now. MSG_1 need to be included after adding such file to open-radar-data repo.

There is also relevant code over in https://github.com/jthielen/xradar/tree/nexrad-level2 by @jthielen, which should be added to this structured reader in a follow-up PR.

### DIFF
--- a/tests/io/test_nexrad_level2.py
+++ b/tests/io/test_nexrad_level2.py
@@ -1,4 +1,3 @@
-
 #!/usr/bin/env python
 # Copyright (c) 2024, openradar developers.
 # Distributed under the MIT License. See LICENSE for more info.
--- a/tests/io/test_nexrad_level2.py
+++ b/tests/io/test_nexrad_level2.py
@@ -56,16 +56,19 @@ def test_open_nexrad_level2_file_ds(nexradlevel2_file):
 
     plt.figure(figsize=(12, 12))
     ds = ds.set_coords("sweep_mode")
-    ds = ds.wrl.georef.georeference()
+    ds = ds.xradar.georeference()
 
     print(ds.DBZH.attrs)
     print(ds.DBZH.encoding)
     print(ds.PHIDP.encoding)
     print(ds.PHIDP.values)
-    ds.DBZH.wrl.vis.plot(ax=221)
-    ds.ZDR.wrl.vis.plot(ax=222)
-    ds.RHOHV.wrl.vis.plot(ax=223)
-    ds.PHIDP.wrl.vis.plot(ax=224)
+    ds.DBZH.plot(x="x", y="y")
+    plt.show()
+    ds.ZDR.plot(x="x", y="y")
+    plt.show()
+    ds.RHOHV.plot(x="x", y="y")
+    plt.show()
+    ds.PHIDP.plot(x="x", y="y")
     plt.show()
 
     # import matplotlib.pyplot as plt
--- a/tests/io/test_nexrad_level2.py
+++ b/tests/io/test_nexrad_level2.py
@@ -5,6 +5,7 @@
 """Tests for `xradar.io.nexrad_archive` module."""
 
 import matplotlib.pyplot as plt
+import pytest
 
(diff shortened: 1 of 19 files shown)


kmuehlbauer:
Finally linting is good. Had outdated version on my machine :grimacing:

kmuehlbauer:
@mgrover1 

This only reads the absolutely necessary metadata to get a hunch on sweeps/moments. Everything else is read when first requested by xarray (so it's lazy too). The reader uses numpy.memmap (like the sigmet reader). Will push the fixture in a minute...

mgrover1:
yes! next week's radar community meeting?

tests/io/test_nexrad_level2.py
```diff
@@ -0,0 +1,353 @@
...
+    for k, v in fh.meta_header.items():
+        print(k, v)
+
+    print("MSG_5", fh.msg_5)
+
+    print(len(fh.data_header))
+
+
+def test_open_nexrad_level2_file_uncompressed():
+    fname_gz = "/home/kai/data/daten/daten/radar_all_over_the_world/NEXRAD/KATX20160601_090111_V06_gz"
```
mgrover1, line 349: @kmuehlbauer - we should have one of these sample files over in open-radar-data?

kmuehlbauer: @mgrover1 Yes, the nice thing is, we can have one with the old uncompressed data and the same file with the bzip2 compressed records. I'll add this next week.

And sorry for breaking the PR here, I'm currently testing with the compression. Seems like I'll need to change the class inheritance, but overall it seems to work.

kmuehlbauer: done

mgrover1 (approved):
This looks great - thanks for your hard work here @kmuehlbauer

mgrover1:
The docs and test coverage look great - we can continue to iterate + test through other PRs - what do you think about merging @kmuehlbauer ?

kmuehlbauer:
Let's wait after the meeting today. There are some issues which I want to show before we get this in.

mgrover1:
I am putting together a notebook right now comparing the data fields from Py-ART + xradar here as well... which should help us understand some of the differences?

kmuehlbauer:
> I am putting together a notebook right now comparing the data fields from Py-ART + xradar here as well... which should help us understand some of the differences?

Great idea! One major difference is that this PR does not mask the data. We would need to think how we want to handle masking in the future. But we can make first make a list with the current differences.

kmuehlbauer:
@mgrover1 If you think we should move forward from here, please go ahead and merge. I'll add several issues dedicated to different aspects/downsides of this implementation. Please also add any issues you find to the tracker, so that we can work on fixing enhancing.

mgrover1:
Great!! Agreed 😄

mgrover1:
@kmuehlbauer - should we close out #40 here and open new issues related to more specific improvements?

kmuehlbauer:
Or let's use this issue (#40) as a tracking issue. We could list all related PR/issues there? Can also pin that issue to the top. WDYT? I'm also good with closing #40.

mgrover1:
Ohh I like that idea... that can be the issue we refer back to + scaffold out. We can rename it to NEXRAD Support, and refer to the different readers and improvements.

mgrover1:
I created #160 to reflect this single reader, and modified #40 accordingly. Sound reasonable @kmuehlbauer ?

PR #172, opened 2024-04-15
Mrr2

rcjackson:
I have added an xarray backend for reading the MRR2 .raw, .pro., and .avg files that are output by the Metek MRR2 software. The usage of the backend is similar to the ones currently implemented in xarray, with the MRR2 engine being 'metek.' In addition. the `xd.io.load_metek_datatree` command will load the MRR2 data into a DataTree.
 
- [x] Tests added


mgrover1:
@rcjackson - looks like some of the tests got hung up - rerunning to see if that is resolved! Thanks for the work so far here 😄

kmuehlbauer:
@rcjackson I've tried to fix this but failed. Somehow the changes in this PR affect the CI so that it breaks for nexrad_leve2 tests. Locally everything works as expected, so not sure how to tackle this problem.

PR #173, opened 2024-04-22
FIX: set fillvalue before applying scale/offset when exporting odim

### LINKED ISSUE #122: xradar and pyart fillvalue discrepancy when working with ODIM
* xradar version: 0.3.0
* Python version: 3.11.4
* Operating System: Windows 11

### Description

I am trying to convert some RAINBOW RB5 data to ODIM for further processing using pyart (the RB5 volumes can't be imported directly due to the variable number of range bins between sweeps not being supported by the pyart package). Anyway, the export and import both work fine in general, however, there is a problem when filling the missing values using the `fillvalue`/`undetect`. Here is a plot obtained from the RB5 volume converted using xradar to ODIM, then loaded and plotted through pyart.

image

The red areas on the plot should be masked, however they show up as areas of high reflectivity.

### What I Did

```
# convert file
datatree = xradar.io.backends.rainbow.open_rainbow_datatree(filepath_start + datetime_str + product + filepath_extension)
xradar.io.export.odim.to_odim(datatree, datetime_str + product + '.h5',source='WMO:11887')

# load new file
radar = pyart.aux_io.read_odim_h5(datetime_str + product + '.h5')
```

Plot showing the unmasked sweep data.
```
plt.imshow(radar.fields['reflectivity_horizontal']['data'])
```
image

### My temporary solution

I have modified the lines around:
https://github.com/openradar/xradar/blob/5d6a3b86b94860aaf966b47eedff24b8b61ed566/xradar/io/export/odim.py#L113C29-L113C29

from
```
val = value.sortby(dim0).values
fillval = _fillvalue * scale_factor
fillval += add_offset
val = (val - add_offset) / scale_factor
val[np.isnan(val)] = fillval
```
to
```
val = value.sortby(dim0).values
val = (val - add_offset) / scale_factor
val[np.isnan(val)] = _fillval
```
as the pyart import seems to expect the exact value that is not transformed to create the mask.

After this change, the plot in pyart seems as expected, with the correct mask:
image

Is this a bug or am I misunderstanding the usage of fillval in the ODIM format?

kmuehlbauer:
- [x] Closes #122
- [ ] Tests added

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -4,6 +4,7 @@
 
 * ADD: Add Alfonso to citation doc ({pull}`169`) by [@mgrover1](https://github.com/mgrover1)
 * ENH: Adding global variables and attributes to iris datatree ({pull}`166`) by [@aladinor](https://github.com/aladinor)
+* FIX: Set fillvalue before applying scale/offset when exporting to odim ({issue}`122`) by [@pavlikp](https://github.com/pavlikp), ({pull}`173`) by [@kmuehlbauer](https://github.com/kmuehlbauer).
 
 ## 0.5.0 (2024-03-28)
 
--- a/xradar/io/export/odim.py
+++ b/xradar/io/export/odim.py
@@ -110,8 +110,8 @@ def _write_odim_dataspace(source, destination, compression, compression_opts):
         val = value.sortby(dim0).values
         fillval = _fillvalue * scale_factor
         fillval += add_offset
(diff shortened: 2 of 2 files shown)


kmuehlbauer:
@mgrover1 If you have any idea how to test this, please let me know. Otherwise this should be good to go.

mgrover1 (approved):
Nice work @kmuehlbauer 😄

PR #197, opened 2024-08-13
FIX: use rstart in meter for ODIM_H5/V2_4

kmuehlbauer:
- [x] Closes #196 
- [ ] Tests added
- [x] Changes are documented in `history.md`

PR #200, opened 2024-08-29
FIX: use dict.get() to retrieve attribute key and return "None" if not available

kmuehlbauer:

This regression was introduced by me in #197.


mgrover1:
Let's get this moving 😊😊

mgrover1 (approved):

PR #202, opened 2024-08-31
ADD: New function - Apply to sweeps

syedhamidali:
- [x] Closes #204 #208 
- [x] Tests added
- [x] Changes are documented in `history.md`

**Description:**

This PR introduces a new utility function, `apply_to_sweeps`, that allows users to apply a custom function to each sweep within a radar volume stored in a `DataTree` object. This is particularly useful for tasks where you need to perform the same operation across multiple sweeps, such as calculating derived quantities like rain rate from radar reflectivity.

**Example Usage:**

Here’s an example of how to use the `apply_to_sweeps` function to calculate rain rates using the Marshall-Palmer power law:

```python
import xradar as xd
from open_radar_data import DATASETS

# Fetch a sample radar file
file = DATASETS.fetch('KATX20130717_195021_V06')

# Open the radar file into a DataTree object
dtree = xd.io.open_nexradlevel2_datatree(file)

# Define the function to calculate rain rate using the Marshall-Palmer relation
def xradar_rain_rate(ds):
    def _rain_rate(dbz, a=200.0, b=1.6):
        """Calculates the inverse of input decibel values
        Convert to rain rate using the Marshall-Palmer relation
        """
        Z = 10.0 ** (dbz / 10.0)  # Convert dBZ to linear Z
        return (Z / a) ** (1.0 / b)  # Apply the Marshall-Palmer power law

    # Apply the rain rate calculation to the DBZH field
    ds['RAIN_RATE'] = _rain_rate(ds['DBZH'])
    ds['RAIN_RATE'].attrs = {'unit': 'mm/h',
                             'long_name': 'Rain Rate'}
    return ds

# Apply the rain rate function to all sweeps in the DataTree
dtree = xd.util.apply_to_sweeps(dtree, xradar_rain_rate)

# Check that the rain_rate variable has been added to one of the sweeps
for var in dtree['sweep_3']:
     print(var)
```
```python
DBZH
VRADH
WRADH
sweep_mode
sweep_number
prt_mode
follow_mode
sweep_fixed_angle
RAIN_RATE
```

Here's another example for example using `filter_radar` function

```python
file = DATASETS.fetch('RAW_NA_000_125_20080411183259')
dtree = xd.io.open_iris_datatree(file)

def filter_radar(ds):
    ds['DBZH_F'] = ds.where((ds['DBZH']>10) & (ds['DBZH']<60))['DBZH']
    return ds

dtree = xd.util.apply_to_sweeps(dtree, filter_radar)
[...]

PR #203, opened 2024-08-31
ADD: map_over_sweeps accessor and decorator

syedhamidali:
- [x] Closes #202 #216 
- [x] Tests added

This PR introduces an apply method to the xradar accessor, allowing users to apply functions to DataTree, Dataset, and DataArray objects seamlessly. This enhancement follows the discussion on generalizing the approach using accessors and aims to provide a more consistent and intuitive API.

### Applying to DataTree

```python
import xradar as xd
from open_radar_data import DATASETS

# Fetch the sample radar file
filename = DATASETS.fetch("sample_sgp_data.nc")

# Open the radar file into a DataTree object
dtree = xd.io.open_cfradial1_datatree(filename)

# Define a function to calculate rain rate from reflectivity
def calculate_rain_rate(ds, ref_field='DBZH'):
    def _rain_rate(dbz, a=200.0, b=1.6):
        Z = 10.0 ** (dbz / 10.0)
        return (Z / a) ** (1.0 / b)

    ds['RAIN_RATE'] = _rain_rate(ds[ref_field])
    ds['RAIN_RATE'].attrs = {'units': 'mm/h', 'long_name': 'Rain Rate'}
    return ds

# Apply the function across all sweeps
dtree = dtree.xradar.apply(calculate_rain_rate, ref_field='corrected_reflectivity_horizontal')

# Inspect the first sweep to see the added RAIN_RATE field
print(dtree['sweep_0'].data_vars)
```
```python
Data variables:
    corrected_reflectivity_horizontal  (azimuth, range) float32 1MB -5.617 .....
    reflectivity_horizontal            (azimuth, range) float32 1MB ...
    RAIN_RATE                          (azimuth, range) float32 1MB 0.01625 ....
```

### Applying to Dataset
```python
# Extract a single sweep as a Dataset
ds = dtree['sweep_0'].to_dataset()

# Apply the function to the Dataset using the xradar accessor
ds = ds.xradar.apply(calculate_rain_rate, ref_field='corrected_reflectivity_horizontal')

# Inspect the Dataset to see the new RAIN_RATE field
print(ds['RAIN_RATE'])
```
```python
<xarray.DataArray 'RAIN_RATE' (azimuth: 400, range: 667)> Size: 1MB
array([[0.01624733, 0.04791909, 0.00855976, ..., 0.02479567, 0.02921897,
               nan],
       [0.01765691, 0.05320087, 0.00698366, ..., 0.01000767,        nan,
               nan]], dtype=float32)
Coordinates:
    time       (azimuth) datetime64[ns] 3kB 2011-05-20T06:42:11.039436300 ......
  * range      (range) float64 5kB 0.0 60.0 120.0 ... 3.99e+04 3.996e+04
  * azimuth    (azimuth) float64 3kB 0.8281 1.719 2.594 ... 358.1 359.0 360.0
    elevation  (azimuth) float64 3kB ...
    latitude   float64 8B ...
    longitude  float64 8B ...
    altitude   float64 8B ...
Attributes:
    units:      mm/h
    long_name:  Rain Rate
```
### Applying to DataArray
```python
# Extract the reflectivity field as a DataArray
da = ds['corrected_reflectivity_horizontal']

# Define a function to add a constant value to the DataArray
def add_constant(da, constant=10):
    return da + constant

# Apply the function using the xradar accessor
da_modified = da.xradar.apply(add_constant, constant=5)

# Inspect the modified DataArray
print(da_modified)
```
```python
<xarray.DataArray 'corrected_reflectivity_horizontal' (azimuth: 400, range: 667)> Size: 1MB
array([[ -0.6171875,   6.8984375,  -5.0703125, ...,   2.3203125,
          3.4609375,         nan],
       [ -0.671875 ,   7.28125  ,  -3.1171875, ...,         nan,
         -8.4765625,         nan]], dtype=float32)
Coordinates:
    time       (azimuth) datetime64[ns] 3kB 2011-05-20T06:42:11.039436300 ......
  * range      (range) float64 5kB 0.0 60.0 120.0 ... 3.99e+04 3.996e+04
  * azimuth    (azimuth) float64 3kB 0.8281 1.719 2.594 ... 358.1 359.0 360.0
    elevation  (azimuth) float64 3kB ...
    latitude   float64 8B ...
    longitude  float64 8B ...
    altitude   float64 8B ...
```


xradar/util.py
```diff
@@ -522,3 +523,43 @@ def get_sweep_keys(dt):
...
+    """
+    for key in list(dtree.children):
+        if "sweep" in key:
+            try:
+                if pass_sweep_name:
+                    dtree[key] = func(dtree[key], sweep=key, *args, **kwargs)
+                else:
+                    dtree[key] = func(dtree[key], *args, **kwargs)
+            except Exception as e:
+                raise RuntimeError(f"An error occurred while processing {key}: {e}")
```
kmuehlbauer, line 563: Maybe re raising like this?

```suggestion
            except Exception as e:
                raise RuntimeError(f"An error occurred while processing {key!r}") from e
```

xradar/accessors.py
```diff
@@ -204,3 +219,25 @@ def add_crs(self) -> dt.DataTree:
         """
         ds = self.xarray_obj
         return ds.pipe(add_crs_tree)
+
+    def apply(self, func, pass_sweep_name=False, *args, **kwargs):
```
kmuehlbauer, line 223: We should keep the signature as clean as possible. We always can extract `pass_sweep_name` from *args or **kwargs.

xradar/accessors.py
```diff
@@ -204,3 +219,25 @@ def add_crs(self) -> dt.DataTree:
...
+            Additional positional arguments to pass to the function.
+        **kwargs : dict
+            Additional keyword arguments to pass to the function.
+
+        Returns
+        -------
+        DataTree
+            The DataTree object after applying the function to all sweeps.
+        """
+        return apply_to_sweeps(self.xarray_obj, func, pass_sweep_name, *args, **kwargs)
```
kmuehlbauer, line 243: Let's handle pass_sweep_name via *args or **kwargs.

xradar/util.py
```diff
@@ -522,3 +523,43 @@ def get_sweep_keys(dt):
...
+    -------
+    DataTree
+        The DataTree object after applying the function to all sweeps.
+
+    Raises
+    ------
+    Exception
+        Re-raises any exception that occurs during the function application.
+    """
+    for key in list(dtree.children):
```
kmuehlbauer, line 555: Here we could extract `pass_sweep_name`.

xradar/accessors.py
```diff
@@ -165,6 +173,13 @@ def get_crs(self):
         radar = self.xarray_obj
         return radar.pipe(get_crs)
 
+    def apply(self, func, *args, **kwargs):
```
kmuehlbauer, line 176: Nice, for DataArray and Dataset this looks like a reimplementation of `.pipe`? The thing is, that we can't use `DataTree.pipe` for our use case. So, yes, let's work with this as suggested.

xradar/accessors.py
kmuehlbauer: I'm wondering if we could simplify this like this:

```python
radar = self.xarray_obj
for key in list(radar.children):
    if "sweep" in key:
        radar[key].ds = func(
            radar[key].to_dataset(), *args, **kwargs)
return radar
```

xradar/accessors.py
```diff
@@ -115,6 +116,13 @@ def get_crs(self):
...
+        """
+        Apply a function to the DataArray object.
+        """
+        da = self.xarray_obj
+        return func(da, *args, **kwargs)
```
kmuehlbauer, line 124: return da.pipe(func, *args, **kwargs)

xradar/accessors.py
```diff
@@ -165,6 +173,13 @@ def get_crs(self):
         radar = self.xarray_obj
         return radar.pipe(get_crs)
 
+    def apply(self, func, *args, **kwargs):
+        """
+        Apply a function to the Dataset object.
+        """
+        ds = self.xarray_obj
+        return func(ds, *args, **kwargs)
```
kmuehlbauer, line 181: We might use the included boilerplate checks within `.pipe`.

```suggestion
        return da.pipe(func, *args, **kwargs)
```

kmuehlbauer (commented):
@syedhamidali Thanks for being that fast. Yes, this is around the lines I was talking before. I've added a couple of comments and suggestion, which we can discuss.

xradar/accessors.py
syedhamidali: Okay, got it.

xradar/accessors.py
syedhamidali: Done!

xradar/accessors.py
syedhamidali: Agreed, I like your suggesstion

xradar/accessors.py
syedhamidali: Yeah, definitely we can.

xradar/accessors.py
syedhamidali: Done!

syedhamidali:
@kmuehlbauer Thanks for these suggestions. Do these changes meet your expectations? I’ve run the methods on some data, and they’re working well. Let me know your thoughts!

docs/history.md
```diff
@@ -1,5 +1,9 @@
 # History
 
+## 0.6.4 (2024-08-31)
```
kmuehlbauer, line 3:
```suggestion
## Development version
```

xradar/accessors.py
```diff
@@ -204,3 +218,28 @@ def add_crs(self) -> dt.DataTree:
...
+        -------
+        DataTree
+            The DataTree object after applying the function to all sweeps.
+        """
+        radar = self.xarray_obj
+        for key in list(radar.children):
+            if "sweep" in key:
+                radar[key].ds = func(radar[key].to_dataset(), *args, **kwargs)
+        return radar
+        # return apply_to_sweeps(self.xarray_obj, func, *args, **kwargs)
```
kmuehlbauer, line 245:
```suggestion
```

xradar/accessors.py
```diff
@@ -204,3 +218,28 @@ def add_crs(self) -> dt.DataTree:
...
+
+        Returns
+        -------
+        DataTree
+            The DataTree object after applying the function to all sweeps.
+        """
+        radar = self.xarray_obj
+        for key in list(radar.children):
+            if "sweep" in key:
+                radar[key].ds = func(radar[key].to_dataset(), *args, **kwargs)
```
kmuehlbauer, line 243: This will overwrite the current tree nodes, as stated in the doc string. Is this really what we want, @syedhamidali? We also could think to construct a new tree here, keeping the original tree as is.

In `georeference()` we also overwrite most likely as we just add the georeferencing x,y,z coordinates. But for the sake of compatibility/comparability of API's it would make sense to return a new DataTree here.

I'd like to get some more opinions, @openradar/xradar WDYT?

aladinor:
what about using we just use [`map_over_subtree`](https://xarray-datatree.readthedocs.io/en/latest/generated/datatree.map_over_subtree.html) functionality already implemented in `xarray.datatree`?

kmuehlbauer:
> what about using we just use [`map_over_subtree`](https://xarray-datatree.readthedocs.io/en/latest/generated/datatree.map_over_subtree.html) functionality already implemented in `xarray.datatree`?

5 vs 16 characters? OK if we count `xradar.apply` it's still 12 vs 16. :grin: 

@aladinor How would we handle this with `map_over_subtree` if we only want to apply this on sweep-nodes?

aladinor:
that's a really good question. We need to make sure it only uses the "sweep" nodes!

kmuehlbauer:
> We need to make sure it only uses the "sweep" nodes!

Yes, this is what we are doing (or want to do) in our boilerplate code. The user should not have to care about that, so we have to take responsibility for that. The only thing which remains, and this is already somehow mis-implemented in the current datatree-accessors, is that the original datatree is overwritten instead of returning a newly created tree.

xradar/accessors.py
aladinor: I agree with @kmuehlbauer. Could we use an `inplace` argument in case users want to overwrite?

I would like recommend this improved version for looping through the `datatree` object. This might be more `datatree` compatible...

```python
for node in radar.match("sweep*").subtree:
    radar[node.path].ds = func(radar[node.path].to_dataset(), *args, **kwargs)
```

kmuehlbauer: @aladinor Your `datatree`-knowledge is really impressive. This looks very neat!

How would your implementation look like to create a new datatree instead of overwriting the sweep-nodes?

kmuehlbauer: > Perhaps, we could use an `inplace` argument if users want to overwrite

IIRC, xarray remove all those `inplace`-implementations from their code base and only return. So maybe we just stick to their behaviour?

aladinor: I was thinking in something like this, 

```python


radar = self.radar
tree = {}
for node in radar.match("sweep*").subtree:
    tree[node.path] = func(radar[node.path].to_dataset(), *args, **kwargs)
return DataTree.from_dict(tree)
```

However, we need to add the other radar_parameters, georeferencing... nodes, as well as the `DataArray`s in the root before returning it.

aladinor: Or copying it and replacing the nodes we want to modify(?)

```python

radar = self.radar.copy()

for node in radar.match("sweep*").subtree:
    radar[node.path] = func(radar[node.path].to_dataset(), *args, **kwargs)
return radar
```

kmuehlbauer: I like the `.from_dict()`-approach as it is also used very much in `datatree` itself. Let's try to figure out a solution going that way.

kmuehlbauer: How about:

```python
# root Dataset
tree = {"/": radar.ds}
# all but root dataset
tree.update({node.path: node for node in radar.subtree if node.path != "/"})
# overwrite sweep nodes
tree.update({node.path: func(radar[node.path].to_dataset(), *args, **kwargs) for node in radar.match("sweep*").subtree if node.path != "/"})
# return new Datatree
return datatree.DataTree.from_dict(tree)
```

aladinor: It looks great! Should we start using `datatree` directly from xarray ?

```python 
from xarray.core.datatree import DataTree
...
```

kmuehlbauer: @aladinor

It is not officially merged yet. So maybe we should keep importing from datatree until this is officially announced. 

You have been attending those meetings, so I'd rely on your expertise if it is safe to move fast here.

syedhamidali:
Hi @kmuehlbauer and @aladinor,

Thanks for your suggestions and for pointing out the node issue, I hadn’t caught that earlier. I’ve made the necessary adjustments and would appreciate your feedback.

mgrover1 (changes requested):
One more suggestion here - can you add an example here @syedhamidali ?

syedhamidali:
## Example

### Applying to DataTree

```python
import xradar as xd
from open_radar_data import DATASETS

# Fetch the sample radar file
filename = DATASETS.fetch("sample_sgp_data.nc")

# Open the radar file into a DataTree object
dtree = xd.io.open_cfradial1_datatree(filename)

# Define a function to calculate rain rate from reflectivity
def calculate_rain_rate(ds, ref_field='DBZH'):
    def _rain_rate(dbz, a=200.0, b=1.6):
        Z = 10.0 ** (dbz / 10.0)
        return (Z / a) ** (1.0 / b)

    ds['RAIN_RATE'] = _rain_rate(ds[ref_field])
    ds['RAIN_RATE'].attrs = {'units': 'mm/h', 'long_name': 'Rain Rate'}
    return ds

# Apply the function across all sweeps
dtree = dtree.xradar.apply(calculate_rain_rate, ref_field='corrected_reflectivity_horizontal')
```

### Applying to Dataset
```python
# Extract a single sweep as a Dataset
ds = dtree['sweep_0'].to_dataset()

# Apply the function to the Dataset using the xradar accessor
ds = ds.xradar.apply(calculate_rain_rate, ref_field='corrected_reflectivity_horizontal')
```

### Applying to DataArray
```python
# Extract the reflectivity field as a DataArray
da = ds['corrected_reflectivity_horizontal']

# Define a function to add a constant value to the DataArray
def add_constant(da, constant=10):
    return da + constant

# Apply the function using the xradar accessor
da_modified = da.xradar.apply(add_constant, constant=5)
```

kmuehlbauer:
@syedhamidali How about making a little notebook of that, which we can refer to in the docs?

This is a very nice feature. Thanks for pushing this forward.

syedhamidali:
> @syedhamidali How about making a little notebook of that, which we can refer to in the docs?
> 
> This is a very nice feature. Thanks for pushing this forward.

@kmuehlbauer Sounds good! Added a notebook, I will update this notebook sometime in the future with more examples.

syedhamidali:
@kmuehlbauer, @mgrover1, let me know if there’s anything else you’d like me to address. I believe these PRs are ready for merging.

kmuehlbauer:
@syedhamidali I think this looks good, but some thoughts of concern:

- `dtree.xradar.apply(func)` - this is really neat, as it works only on sweep nodes
- `ds.xradar.apply(func)` - isn't this more or less the same as `ds.pipe(func)`? I do not really see any additional benefit here
- `da.xradar.apply(func)` - same as above

I do not think we should add another function which does actually the same as `.pipe`.

Also I'm not sure if `.apply` is the best naming in this context. Over in xarray `map_over_subtree` was changed in `map_over_datasets` for ruling out any misunderstandings. Maybe `map_over_datasets` can be parameterized like name="sweep" where "sweep" is a string which the node-name should contain. Then we could just let xarray do the hard work.

Pinging also @aladinor and @mgrover1 for input.

aladinor:
Thanks, @kmuehlbauer, for sharing your thoughts.  

>  Maybe `map_over_datasets` can be parameterized like name="sweep" where "sweep" is a string that the node-name should contain. Then we could just let xarray do the hard work.

I think @syedhamidali did a great job with the apply function. However, as you mentioned, this function might overlap with `map_over_datasets`. This could be also an approach to this idea

```python 
# opening datatree stored in zarr
dtree = xr.backends.api.open_datatree(
  file, 
  engine='zarr', 
  consolidated=True, 
  chunks={}
)

# we can use the `match` method to get only the sweeps.`
dt_sweeps = dtree.match("sweep_*")

# and then map_over_datasets
dt_sweeps.map_over_datasets(fucn)

```
And as @kmuehlbauer mentioned, we can just let Xarray do the heavy lifting. However, I might be missing something like if we want to create a wrapper within `xradar`. In that case, we might need to rebuild the `datatree` by incorporating all the other nodes not considered in the `map_over_datasets`.

Please let me know your thoughts.

kmuehlbauer:
@syedhamidali Not to be mistaken or misunderstood, It's good to see such excellent contributions coming in.

What I mean't and with the possibilities xarray gives us, we might get some really neat functionality into xradar without doing that much and keeping our maintenance burden low.

How about something like this (works with latest xarray-main and xradar-`xarray-nightly` branch :


```python
import xradar as xd
from open_radar_data import DATASETS

# Fetch the sample radar file
filename = DATASETS.fetch("sample_sgp_data.nc")

# Open the radar file into a DataTree object
dtree = xd.io.open_cfradial1_datatree(filename)


# define the xradar decorator to just work on sweep nodes 
# and let other nodes out unchanged
# thats quick'ndirty, so the code might not be up to standards
from functools import wraps

def map_over_sweeps(func):

    @wraps(func)
    def wrapper_func(*args, **kwargs):
        if "range" in args[0].dims:
            return func(*args, **kwargs)
        else:
            return args[0]
    return wrapper_func


# Define a function to calculate rain rate from reflectivity
# apply the decorators

@xr.map_over_datasets
@map_over_sweeps
def calculate_rain_rate(ds, ref_field='DBZH'):
    def _rain_rate(dbz, a=200.0, b=1.6):
        Z = 10.0 ** (dbz / 10.0)
        return (Z / a) ** (1.0 / b)

    ds = ds.assign(RAIN_RATE=_rain_rate(ds[ref_field]))
    ds.RAIN_RATE.attrs = {'units': 'mm/h', 'long_name': 'Rain Rate'}
    return ds

# Apply the function across all sweeps
dtree = dtree.pipe(calculate_rain_rate, ref_field='corrected_reflectivity_horizontal')

# Inspect the first sweep to see the added RAIN_RATE field
print(dtree['sweep_0'].data_vars)
``` 

So finally the only function we have to maintain is the `map_over_sweeps`-decorator. I think it should be possible to even hide the `xr.map_over_datasets`-decorator, doing the coding of the wrapper functions the right way.

syedhamidali:
> @syedhamidali I think this looks good, but some thoughts of concern:
> 
> * `dtree.xradar.apply(func)` - this is really neat, as it works only on sweep nodes
> * `ds.xradar.apply(func)` - isn't this more or less the same as `ds.pipe(func)`? I do not really see any additional benefit here
> * `da.xradar.apply(func)` - same as above
> 
> I do not think we should add another function which does actually the same as `.pipe`.
> 
> Also I'm not sure if `.apply` is the best naming in this context. Over in xarray `map_over_subtree` was changed in `map_over_datasets` for ruling out any misunderstandings. Maybe `map_over_datasets` can be parameterized like name="sweep" where "sweep" is a string which the node-name should contain. Then we could just let xarray do the hard work.
> 
> Pinging also @aladinor and @mgrover1 for input.

Thanks, @kmuehlbauer! You’re right—ds.pipe(func) and da.pipe(func) already cover what .apply does, so no need to add redundancy.

syedhamidali:
Thanks, @aladinor! I agree that parameterizing map_over_datasets for sweeps is a cleaner approach. If we decide to create a wrapper for xradar, handling non-sweep nodes carefully will be key. Appreciate your thoughts!

I gave it a try, something like this, but there are a lot of issues with that, maybe you can take a look

```python
def apply(self, func, *args, **kwargs):
    radar = self.xarray_obj

    # Apply the function only to the sweep nodes
    dt_sweeps = radar.match("sweep_*")
    modified_sweeps = dt_sweeps.map_over_datasets(func, *args, **kwargs)

    # Rebuild the DataTree, including non-sweep nodes
    new_tree = {"/": radar.ds}
    new_tree.update({node.path: node.ds for node in radar.subtree if node.path != "/"})
    new_tree.update({node.path: modified_sweeps[node.path].to_dataset() for node in modified_sweeps.subtree})

    return dt.DataTree.from_dict(new_tree)
```

```bash
AttributeError: Mutation of the DatasetView is not allowed, please use `.__setitem__` on the wrapping DataTree node, or use `dt.to_dataset()` if you want a mutable dataset. If calling this from within `map_over_datasets`,use `.copy()` first to get a mutable version of the input dataset.
Raised whilst mapping function over node with path /sweep_0
```

Now, tried to fix this error

```python
def apply(self, func, *args, **kwargs):
    """    """

    radar = self.xarray_obj  # Get the DataTree object

    # Match only the sweep nodes (e.g., sweep_0, sweep_1, etc.)
    dt_sweeps = radar.match("sweep_*")

    # Apply the function to all sweep datasets, ensuring each dataset is copied before modification
    def apply_func(ds, *args, **kwargs):
        return func(ds.copy(), *args, **kwargs)  # Copy the dataset before applying the function

    # Apply the function across the matched sweep nodes using map_over_datasets
    modified_sweeps = dt_sweeps.map_over_datasets(apply_func, *args, **kwargs)

    # Rebuild the DataTree, including non-sweep nodes (like root nodes) if necessary
    tree_dict = {"/": radar.ds}  # Add the root dataset ("/" path)
    tree_dict.update({node.path: node.ds for node in radar.subtree if node.path != "/"})  # Add non-sweep nodes
    tree_dict.update({node.path: modified_sweeps[node.path] for node in modified_sweeps.subtree})

    # Return the updated DataTree
    return dt.DataTree.from_dict(tree_dict)
```
```bash
KeyError: 'Already a node object at path /sweep_0'
```

syedhamidali:
Thanks again, @kmuehlbauer! I like the idea of using map_over_sweeps to focus on sweep nodes, keeping maintenance light. The decorator approach works well. What’s your take on encouraging users to rely on decorators for applying functions? Seems efficient, but open to your thoughts!

This seems to work fine with the new xarray-datatree

```python
def map_over_sweeps(self, func, *args, **kwargs):
    """    """
    radar = self.xarray_obj

    # Create a new tree dictionary
    tree = {"/": radar.ds}  # Start with the root Dataset

    # Add all nodes except the root
    tree.update({node.path: node.ds for node in radar.subtree if node.path != "/"})

    # Apply the function to all sweep nodes and update the tree dictionary
    tree.update(
        {
            node.path: func(radar[node.path].to_dataset(), *args, **kwargs)
            for node in radar.match("sweep*").subtree
            if node.path.startswith("/sweep")
        }
    )

    # Return a new DataTree constructed from the modified tree dictionary
    return xr.DataTree.from_dict(tree)

```

kmuehlbauer:
> The decorator approach works well. What’s your take on encouraging users to rely on decorators for applying functions? Seems efficient, but open to your thoughts!

OK, so from my perspective xarray is already encouraging the use of decorators via [accessors](https://docs.xarray.dev/en/latest/internals/extending-xarray.html). With the next xarray version the [`map_over_datasets`-decorator](https://docs.xarray.dev/en/latest/generated/xarray.map_over_datasets.html#xarray.map_over_datasets) is first introduced.

We could imagine following that path by introducing a `map_over_sweeps`-decorator. And I agree that such `map_over_sweeps`-decorator should be backed by a `dtree.xradar.map_over_sweeps` accessor based functionality as you suggested above. I'd try to align these two features as much as possible. So I'd try to re-use the decorator in the accessor-function. That means, instead of reimplementing the distribution of `func` to the sweeps and rebuilding the tree by ourselves I'd rely on xarray machinery and just do the selection of the sweeps inside the decorator. Does that make sense? BTW, xarray does the same with the implementation of [`map_over_datasets`function](https://docs.xarray.dev/en/latest/generated/xarray.DataTree.map_over_datasets.html#xarray.DataTree.map_over_datasets).

kmuehlbauer:
@syedhamidali Your above example (https://github.com/openradar/xradar/pull/203#issuecomment-2417513008) would then look like:

```python

# define the xradar decorator to just work on sweep nodes 
# and let other nodes out unchanged
# thats quick'ndirty, so the code might not be up to standards
import functools
import xarray as xr

def map_over_sweeps(func):
    
    @functools.wraps(func)
    def _func(*args, **kwargs):
        # we do not have access to any names or other DataTree specifics
        # so we need to rely on the layout of the given dataset
        # maybe we can add something like `is_sweep` to check for that?
        if "range" in args[0].dims:
            return func(*args, **kwargs)
        else:
            return args

    # we need another wrap here for the xarray decorator
    @xr.map_over_datasets
    def _map_over_sweeps(*args, **kwargs):
        return _func(*args, **kwargs)
            
    return _map_over_sweeps

# defined inside XarrayAccessor
def map_over_sweeps(self, func, *args, **kwargs):
    """    """
    # add our decorator for input function
    @map_over_sweeps
    def _func(*args, **kwargs):
        return func(*args, **kwargs)
    
    return self.xarray_obj.pipe(_func, *args, **kwargs)
```

The user would then have two options, one via accessor and one via decorator. There are surely use cases for both. The good thing is, that both ways are implemented solely by our and the xarray decorator and we only use xarray machinery (beside our little "range"-hack).

```python
import xradar as xd
from open_radar_data import DATASETS

# Fetch the sample radar file
filename = DATASETS.fetch("sample_sgp_data.nc")

# Open the radar file into a DataTree object
dtree = xd.io.open_cfradial1_datatree(filename)

# without decorator
def calculate_rain_rate(ds, ref_field='DBZH'):
    def _rain_rate(dbz, a=200.0, b=1.6):
        Z = 10.0 ** (dbz / 10.0)
        return (Z / a) ** (1.0 / b)

    ds = ds.assign(RAIN_RATE=_rain_rate(ds[ref_field]))
    ds.RAIN_RATE.attrs = {'units': 'mm/h', 'long_name': 'Rain Rate'}
    return ds

# with decorator
@map_over_sweeps
def calculate_rain_rate2(ds, ref_field='DBZH'):
    def _rain_rate(dbz, a=200.0, b=1.6):
        Z = 10.0 ** (dbz / 10.0)
        return (Z / a) ** (1.0 / b)

    ds = ds.assign(RAIN_RATE=_rain_rate(ds[ref_field]))
    ds.RAIN_RATE.attrs = {'units': 'mm/h', 'long_name': 'Rain Rate'}
    return ds

# invocation via accessor
dtree2 = dtree.xradar.map_over_sweeps(calculate_rain_rate, ref_field='corrected_reflectivity_horizontal')
display(dtree2["sweep_0"])

# invocation via decorator + pipe
dtree3 = dtree.pipe(calculate_rain_rate2, ref_field='corrected_reflectivity_horizontal')
display(dtree3["sweep_0"])
```

syedhamidali:
@kmuehlbauer I've one quick question, should I also add the `from .accessors import map_over_sweeps  # noqa` to the `xradar/init.py` so that we can call it `@xd.map_over_sweeps`?

kmuehlbauer:
@syedhamidali 

> I've one quick question, should I also add the `from .accessors import map_over_sweeps # noqa` to the `xradar/__init__.py` so that we can call it `@xd.map_over_sweeps`?

Yes, It would be great to have the decorator available directly from the module, instead importing from submodule.

syedhamidali:
@kmuehlbauer, okay, I am on it, thanks!

syedhamidali:
@kmuehlbauer It should work when you merge the nightly branch

kmuehlbauer:
@syedhamidali Great! This looks really nice, thanks! The nightly run is green, so it works! Let's wait until the new xarray version is out, or we just merge it into the feature branch. Not sure, what's less invasive.

kmuehlbauer:
@syedhamidali Unfortunately xarray has removed the decorator for map_over_datasets. So this will need some fixing. I'll try to fix it today :grimacing:

kmuehlbauer:
@syedhamidali I've fixed the decorator issue, tweaked one test and fixed some docs/notebooks.

Please have another look, if you are good with the PR. Then we should merge.

syedhamidali:
Hi @kmuehlbauer, Thank you for handling the decorator issue and the additional tweaks! Everything looks great on my end after reviewing the changes, so I’m good with the PR. I appreciate your support with this and the updates to the docs/notebooks. Let me know if there’s anything else you’d like me to address.
Thanks again!

kmuehlbauer:
Great, then let's get this in! Thanks for sticking to this during the datatree conversion process.

kmuehlbauer (approved):

aladinor (approved):

xradar/accessors.py
```diff
@@ -204,3 +218,41 @@ def add_crs(self) -> dt.DataTree:
...
+
+        Returns
+        -------
+        DataTree
+            A new DataTree object with the function applied to all sweeps.
+        """
+        radar = self.xarray_obj
+
+        # Create a new tree dictionary
+        tree = {"/": radar.ds}  # Start with the root Dataset
```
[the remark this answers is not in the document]
kmuehlbauer: @aladinor As you've spotted this, would it be possible to add a fix in your from_dict PR?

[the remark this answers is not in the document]
kmuehlbauer: @aladinor, is this still relevant? I think, the accessor (now: `map_over_sweeps`) was rewritten using xarray machinery (`map_over_datasets`).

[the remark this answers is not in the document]
kmuehlbauer: Nevermind @aladinor. Somehow your old comment was presented as new in the UI. :grimacing;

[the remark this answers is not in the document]
aladinor: Sorry @kmuehlbauer I approved an old review by mistake. I hope this does affect the PR :grimacing;

PR #213, opened 2024-10-10
Add xarray nightly CI run

### LINKED ISSUE #149: Migrate datatree imports to xarray
* xradar version: dev
* Python version: All
* Operating System: All

### Description

With datatree being deprecated soon (moving into the core xarray library), we need to change any imports to pull from xarray instead of the xarray-datatree package

https://github.com/xarray-contrib/datatree

### What I Did

I saw this on the xarray-datatree readme.

kmuehlbauer:
- [x] Closes #149
- [x] Tests added/changes
- [ ] Docs changed accordingly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,50 @@ jobs:
         fail_ci_if_error: false
         token: ${{ secrets.CODECOV_TOKEN}}
 
+  build_2:
+    name: xarray nightly tests - linux
+    runs-on: ubuntu-latest
+    needs: [ lint ]
+    continue-on-error: true
+    defaults:
+      run:
+        shell: bash -l {0}
+    env:
+      CONDA_ENV_FILE: ci/unittests.yml
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.12" ]
+        numpy-version: [ "2" ]
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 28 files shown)


kmuehlbauer:
As can be seen from the nightly CI run, xarray has already removed things from it's codebase in main branch and breaks the original datatree. So we'll have to make a move now.

kmuehlbauer:
Yikes! The nightly-xarray runs works with the changes from this PR. I've adapted `DataTree` related ode and tests. Also used context manager in the tests as far as possible. I think I've (hopefully) identified the bugging bus-error issue in the CI. It is resolved by not running some tests in parallel (`loadscope` instead `loadfile`, see https://pytest-xdist.readthedocs.io/en/latest/distribution.html#running-tests-across-multiple-cpus for the different meaning).

@openradar/xradar How are your feelings, should we abandon the old [datatree](https://github.com/xarray-contrib/datatree) (which will be archived soon)? 

This would mean, that we will have to pin xradar to the next xarray version (eg. 2024-10) for our next release. At least I do not see a good thing in maintaining two different implementations. At the same time, all normal tests will return green, if next xarray version is available.

aladinor:
Thanks, @kmuehlbauer, for bringing this up. Indeed, the `xr.DataTree` will be released soon, I think with the next release of `xarray`. So, let's move towards using the `xarray` mainstream. I'd be happy to help!

PR #214, opened 2024-10-16
FIX: testsuite

kmuehlbauer:
… should finally resolve annoying bus error in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
         python -c "import xradar; print(xradar.version.version)"
     - name: Test with pytest
       run: |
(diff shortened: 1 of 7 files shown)


kmuehlbauer:
This is frustrating. It seems that the nightly xarray test over in #213 works now. But there are more changes there wrt contextmanager. So I'll pull these changes here, too.

kmuehlbauer:
When running all tests in serial fashion, no bus error. There is something very strange going on with numpy.memmap and how we use the file fixtures for nexrad level2 files. 

I've removed one test (which is already in test_io.py) from nexrad_level2 tests and reinstated the --dist loadfile (runs all tests in one file in serial fashion).

PR #219, opened 2024-10-25
Pin xarray to 2024.9.0 and release v0.7.0

kmuehlbauer:
- [ ] Closes #xxxx
- [ ] Tests added
- [x] Changes are documented in `history.md`

PR #220, opened 2024-10-25
Creating datatree from dict

aladinor:
Hi @kmuehlbauer, 

Based on `xarray.Datatree`, they used `from_dict` to build the tree. Should we move towards using something similar to this draft?

- [ ] Closes #xxxx
- [ ] Tests added


kmuehlbauer:
Sure, can we add this on top of #219?

PR #224, opened 2024-10-28
ADD: Transform function CF1 to CF2 and vice versa

syedhamidali:
- [x] Closes #136  
- [x] Tests added

### DIFF
--- a/examples/notebooks/Transform.ipynb
+++ b/examples/notebooks/Transform.ipynb
@@ -20,8 +20,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import xradar as xd\n",
-    "from open_radar_data import DATASETS"
+    "from open_radar_data import DATASETS\n",
+    "\n",
+    "import xradar as xd"
    ]
   },
   {
--- a/examples/notebooks/Transform.ipynb
+++ b/examples/notebooks/Transform.ipynb
@@ -1,0 +1,104 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Transform"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Imports"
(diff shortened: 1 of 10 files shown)


examples/notebooks/Transform.ipynb
```diff
@@ -0,0 +1,104 @@
...
+    "file = DATASETS.fetch(\"cfrad.20080604_002217_000_SPOL_v36_SUR.nc\")\n",
+    "dtree = xd.io.open_cfradial1_datatree(file)\n",
+    "dtree"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Transform CF1 to CF2"
```
kmuehlbauer, line 50: Isn't this the other way round?

examples/notebooks/Transform.ipynb
```diff
@@ -0,0 +1,104 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "file = DATASETS.fetch(\"cfrad.20080604_002217_000_SPOL_v36_SUR.nc\")\n",
+    "dtree = xd.io.open_cfradial1_datatree(file)\n",
```
kmuehlbauer, line 42: You might read it as pure netcdf (xr.open_dataset) and the transform to CF2 and back to CF1. But not necessary.

kmuehlbauer (approved):
LGTM, Slight inconsistency in the notebook. But after that we can merge, I think.

examples/notebooks/Transform.ipynb
syedhamidali: You're right :)

kmuehlbauer:
@syedhamidali Thanks for coming back to this. Great work!

syedhamidali:
You’re my inspiration, @kmuehlbauer. None of this would have been possible without your support every step of the way.

Issue #163, opened 2024-03-15
ENH: Bring in extended list of supported message structures from MetPy for Level2 reader


jthielen:
As referenced in https://github.com/openradar/xradar/pull/158 and discussed with @mgrover1 in the MetPy telecon yesterday, there is still some relevant code in https://github.com/openradar/xradar/compare/main...jthielen:xradar:nexrad-level2 that may be helpful, namely that of a reconciled list of message structures combining those from both Py-ART's and MetPy's. Given that prior effort used MetPy's bespoke `NamedStruct` and `BitField` objects to handle this, there would likely be a quite a bit of refactoring needed (going either direction..modifying the reconciled lists to match the current xradar approach, or updating the xradar approach to use these MetPy-style objects). This issue can hopefully track this particular facet of the overall effort of https://github.com/openradar/xradar/issues/40.

kmuehlbauer:
Thanks @jthielen for pushing this here.

I can have a look what would be necessary to incorporate things like `BitField` into the current structure. 

I'm also interested to check if and how `NamedStruct` can be utilized for the current processing.

wx4stg:
I'll just re-up this as there's a particular proprietary radar file that claims to be "level 2 format" (I can't share sadly, and I don't have time to look into it right now... I'm being soooo helpful, I know :/ ) that opens in metpy and pyart but does not open in xradar...

PR #231, opened 2024-11-02
FIX: 8bit/16bit, big-endian/little-endian in nexrad reader

### LINKED ISSUE #230: Issue with ZDR when reading data with nexrad_level2 reader
* xradar version: 0.7
* Python version: 3.12
* Operating System: MacOS

### Description

I think the nexrad_level2 reader is parsing wrong values for ZDR

### What I Did

```python

import os
import pytz
import logging
import fsspec

def get_nexrad(timezone, date, site, save_dir, localize=False):
    """
    Download NEXRAD radar data from an S3 bucket and save it to a directory.

    Parameters
    ----------
    timezone : str
        The timezone of the input `date`.
    date : datetime.datetime
        The date and time to download the data for, either in UTC or local time.
    site : str
        The radar site code (e.g., 'KGWX').
    save_dir : str
        The directory where downloaded files should be saved.
    localize : bool, optional
        If True, interprets `date` as local time in the specified `timezone`.
        Otherwise, `date` is assumed to be in UTC. Default is False.
    """
    try:
        # Convert date to UTC if needed
        utc_date = (
            pytz.timezone(timezone).localize(date).astimezone(pytz.utc)
            if localize
            else date
        )
        logging.info(f"Time: {utc_date}")

        # Initialize the S3 filesystem with anonymous access
        fs = fsspec.filesystem("s3", anon=True)
        nexrad_path = utc_date.strftime(
            f"s3://noaa-nexrad-level2/%Y/%m/%d/{site}/{site}%Y%m%d_%H*"
        )
        files = sorted(fs.glob(nexrad_path))
        
        # Ensure the save directory exists
        os.makedirs(save_dir, exist_ok=True)

        # Download and save files, excluding metadata files (_MDM)
        downloaded_files = []
        for file in files:
            if not file.endswith("_MDM"):
                local_path = os.path.join(save_dir, os.path.basename(file))
                fs.get(file, local_path)
                downloaded_files.append(local_path)

        return downloaded_files
    except Exception as e:
        logging.error("Error in processing: %s", e)
        return []

# Load NEXRAD data from S3 Bucket
site = "KTLH"
timezone = "UTC"
date = datetime(2024, 9, 26, 23)

radar_files = get_nexrad(timezone, date, site, save_dir=f"{site}_HELENE",)
radar_files

import pyart
import xradar as xd

dtree = xd.io.open_nexradlevel2_datatree('KTLH_HELENE/KTLH20240926_233515_V06')

dtree = dtree.xradar.georeference()

dtree['sweep_0']['ZDR'].plot(x='x', y='y')
```
image

```python
ds = dtree['sweep_2'].to_dataset()

ds = ds.where(ds.DBZH >= 0)

ds['ZDR'].plot(x='x', y='y')
```
image

```python
radar = pyart.io.read_nexrad_archive('KTLH_HELENE/KTLH20240926_233515_V06')

display = pyart.graph.RadarDisplay(radar)

display.plot("differential_reflectivity", )

```

image

kmuehlbauer:
- [x] Closes #230 
- [ ] Tests added


kmuehlbauer:
@syedhamidali Could you please test your workflow with these changes. 

We would have to create a release branch for the 0.7-line and backport this. Otherwise users are forced to upgrade to latest xarray. This might take some time to figure out the workflows, though.

xradar/io/backends/nexrad_level2.py
```diff
@@ -1253,8 +1252,11 @@ def _getitem(self, key):
         except KeyError:
             self.datastore.root.get_data(self.group, self.name)
             data = self.datastore.ds["sweep_data"][self.name]["data"]
+        # see 3.2.4.17.6 Table XVII-I Data Moment Characteristics and Conversion for Data Names
         if self.name == "PHI":
```
kmuehlbauer, line 1256: This actually only represents the latest ICD's https://www.roc.noaa.gov/interface-control-documents.php. How do we check for data based on earlier ICD's? We could just add a table with the versions and the differences.



We would also need to mask out 0 (indicates received signal is below threshold) and 1 (
indicates range folded data).

syedhamidali:
@kmuehlbauer Yes looks better now. Thank you!
image

xradar/io/backends/nexrad_level2.py
kmuehlbauer: @syedhamidali This will only read PHI and ZDR correctly for the latest ICD based radar data. I'll add a check for word_size for the ZDR to only use 11 bit for 2 byte data. Any other suggestions from your side, how to tackle this properly?

syedhamidali: I think it reads as it was before

image

kmuehlbauer: OK, I meant, if we have older data with 1byte ZDR, that might not read correct.

kmuehlbauer: When you are on it, could you check if PHIDP does make sense?

syedhamidali: Looks good to me

image

syedhamidali: and this looks also fine

image

kmuehlbauer: OK, great! Thanks for checking. I think with the current 2byte check, we are on the safe side. All other enhancements/fixes for nexrad should be tackled when working on #40 and #163.

PR #252, opened 2024-12-06
FIX datatree reader now works with sweeps containing different variables

egouden:
This is a quick fix to allow reading radar data containing different variables from sweep to sweep.

I would like to share my experience with contributing. Running the tests failed after following the instructions. I had to do a significant analysis and several test runs. I am not an expert in packaging but I suggest to add the following python packages to the requirement list: 

open_radar_data
jupyter
cartopy
boto3
s3fs
zarr

_ffmpeg_ is also required

### DIFF
--- a/xradar/io/backends/common.py
+++ b/xradar/io/backends/common.py
@@ -240,7 +240,7 @@ def _get_required_root_dataset(ls_ds, optional=True):
         remove_root ^= set(optional_root_vars)
     remove_root ^= {"sweep_number", "fixed_angle"}
     remove_root &= data_var
-    root = [sweep.drop_vars(remove_root) for sweep in ls_ds]
+    root = [sweep.drop_vars(remove_root, errors="ignore") for sweep in ls_ds]
     root_vars = {x for xs in [sweep.variables.keys() for sweep in root] for x in xs}
     # rename variables
     # todo: find a more easy method not iterating over all variables


kmuehlbauer:
Thanks @egouden, can you share a repr of a file (or the file itself?) where this breaks? It would be great if we could somehow test this, or maybe also parametrize this (eg by kwarg).

kmuehlbauer:
> I am not an expert in packaging but I suggest to add the following python packages to the requirement list:
> 
> open_radar_data jupyter cartopy boto3 s3fs zarr
> 
> _ffmpeg_ is also required

It would make sense to open a separate issue on how to deal with additional dependencies.

egouden:
> Thanks @egouden, can you share a repr of a file (or the file itself?) where this breaks? It would be great if we could somehow test this, or maybe also parametrize this (eg by kwarg).

Thanks for the quick reply. An example file can be found [here][(http://s3-eu-west-1.amazonaws.com/fmi-opendata-radar-volume-hdf5/2023/07/08/fianj/202307081500_fianj_PVOL.h5].

xradar/io/backends/common.py
```diff
@@ -240,7 +240,7 @@ def _get_required_root_dataset(ls_ds, optional=True):
         remove_root ^= set(optional_root_vars)
     remove_root ^= {"sweep_number", "fixed_angle"}
     remove_root &= data_var
-    root = [sweep.drop_vars(remove_root) for sweep in ls_ds]
+    root = [sweep.drop_vars(remove_root, errors="ignore") for sweep in ls_ds]
```
kmuehlbauer, line 243:
```suggestion
    # ignore errors for variables which exist in one sweep but not the other
    root = [sweep.drop_vars(remove_root, errors="ignore") for sweep in ls_ds]
```

kmuehlbauer:
I wonder why this didn't surface earlier. The proposed fix is working as intended and should also be the canonical way of doing. No tests needed as far as I'm concerned.

Thanks @egouden!

PR #255, opened 2024-12-17
ENH: keep selected variables in dataset/datatree

egouden:
It is often not needed or inefficient to work with all observed variables contained in a radar sweep or radar volume. 

It is possible to select observed variables directly but the ancillary and metadata variables are lost.

This enhancement provides an easy selection of observed variables while keeping the data model.


egouden:
A new version has been submitted to enhance clarity.

Issue #104, opened 2023-04-03
Distinction between dataset variables and ancillary variables


egouden:
I cannot find a way to distinguish between the dataset variables and metadata variables.

kmuehlbauer:
@egouden Could you please elaborate along an example? It's hard to guess what you are referring to.

egouden:
What about using attributes to store these ancillary variables?

kmuehlbauer:
Are you referring to those ancillary variables https://docs.openradarscience.org/projects/xradar/en/stable/datamodel.html#ancillary-variables?

It might not be possible as those variables are data variables in the underlying netcdf file. 

Maybe can describe the use case you have in mind and why you need this separated?

egouden:
There are actually 3 types of sweep variables:
- observed variables
- ancillary variables related to one or more observed variables
- metadata variables related to the sweep itself

Accessing only the observed variables should be a basic feature indeed. Is it not already needed in [our code](https://github.com/openradar/xradar/blob/main/xradar/io/export/odim.py#L80)?

egouden:
This has been solved by https://github.com/openradar/xradar/pull/257

PR #257, opened 2025-01-28
ADD: function to select dataset variables in sweep

egouden:
The WMO conventions define two types of data variables for the sweep object:
- dataset variables, with dimension of time and range (e.g. DBZH, RR, quality1, ...)
- metadata variables, typically a scalar (e.g. sweep_number, sweep_mode, polarization_mode, ...)

These variables can be accessed together by the _data_vars_ property of the sweep object (xarray Dataset). However it is not possible to make the distinction between the two types of variables. Two new functions allow to access these variables separately.

It is possible to select dataset variables directly using sweep[list of variables]. But this results in the loss of the metadata. A new function allows to select dataset variables while keeping metadata variables.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -1,7 +1,10 @@
 # History
 
 ## Development Version
-
+* ADD: function to select dataset variables in sweep ({issue}`104`) ({pull}`254`) by [@egouden](https://github.com/egouden)
+* ADD: function to get dataset variables in sweep
+* ADD: function to get metadata variables in sweep
+* FIX: typo in accessors module: Dataarray -> Dataset
 * FIX: Update missing deps for virtualenv environments via "requirements_dev.txt". ({issue}`253`) ({pull}`274`) by [@Steve-Roderick](https://github.com/Steve-Roderick).
 
 ## 0.9.0 (2025-02-07)
--- a/xradar/accessors.py
+++ b/xradar/accessors.py
@@ -120,7 +120,7 @@ class XradarDataArrayAccessor(XradarAccessor):
 
 @xr.register_dataset_accessor("xradar")
 class XradarDataSetAccessor(XradarAccessor):
-    """Adds a number of xradar specific methods to xarray.DataArray objects."""
+    """Adds a number of xradar specific methods to xarray.Dataset objects."""
 
     def georeference(
         self, earth_radius=None, effective_radius_fraction=None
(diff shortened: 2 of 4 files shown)


xradar/accessors.py
```diff
@@ -120,7 +120,7 @@ def get_crs(self):
 
 @xr.register_dataset_accessor("xradar")
 class XradarDataSetAccessor(XradarAccessor):
-    """Adds a number of xradar specific methods to xarray.DataArray objects."""
+    """Adds a number of xradar specific methods to xarray.Dataset objects."""
```
kmuehlbauer, line 123: Nice catch!

xradar/util.py
```diff
@@ -648,3 +650,55 @@ def _map_over_sweeps(*args, **kwargs):
         return xr.map_over_datasets(functools.partial(_func, **kwargs), *args)
 
     return _map_over_sweeps
+
+
+def dataset_vars(sweep):
```
kmuehlbauer, line 655: I'm worrying a bit about the naming. This is too close to the xarray nomenclature (`Dataset.data_vars`). I do not have an ad-hoc suggestion, maybe prepending it with something suitable?

kmuehlbauer (commented):
Thanks @egouden! That would be a nice addition. We need to have a broader discussion on appropriate naming.

xradar/util.py
egouden: I suggest "get_".

egouden:
This is related to #104.

A probably better solution in the future would be to separate data and metadata in two different datasets inside the sweep group.

kmuehlbauer:
@egouden Can you fix the merge conflicts or rebase on main? Otherwise this seems good to go.

docs/history.md
```diff
@@ -1,7 +1,10 @@
 # History
 
 ## Development Version
-
+* ADD: function to select dataset variables in sweep
```
kmuehlbauer, line 4: Please link related issues and the PR.

kmuehlbauer (approved):
Thanks @egouden! Just one last suggestion, after that we can merge.

egouden:
Is this ready?

kmuehlbauer:
Ready from my side. @mgrover1 would you mind having a last look here?

mgrover1:
Absolutely!

> Ready from my side. @mgrover1 would you mind having a last look here?

mgrover1 (approved):
Nice addition - thanks for adding tests here and updating the history file! Good to go!

syedhamidali:
Can we edit some notebook and implement an example of this function somewhere?

PR #280, opened 2025-06-27
Nexrad streaming

### LINKED ISSUE #265: Nexrad reader does not support .gz files neigther data streaming
Tracked also in #258 

## Xradar should support  compress files as well as data streaming

`pyart` library allows to open in different formats including `Bz`, `gz` and files stored in an s3 bucket (data) streaming. I think we might want to implement a similar solution to xradar reader. 

- Create/adapt [prepare_for_read](https://arm-doe.github.io/pyart/_modules/pyart/io/common.html#prepare_for_read)
- Adapt xradar backends to support data streaming as in `PyArt`

aladinor:
- [x] Closes #265 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,7 @@ jobs:
             h5py
             lat_lon_parser
             netCDF4
-            open-radar-data>=0.3.0
+            open-radar-data>=0.4.2
             packaging
             pandas
             pip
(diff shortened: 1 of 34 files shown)


kmuehlbauer (approved):

aladinor:
now sure. I might need to double-check

mgrover1 (approved):

PR #281, opened 2025-07-01
use fixture for making temp file

egouden:
This builds on 277.


tests/conftest.py
```diff
@@ -160,3 +161,12 @@ def metek_raw_gz_file(tmp_path_factory):
         with open(fnameo, "wb") as fout:
             shutil.copyfileobj(fin, fout)
     return fnameo
+
+
+@pytest.fixture
```
kmuehlbauer, line 166: Nice @egouden, would it make sense to create a second fixture which directly creates the name (for single file use case)?

```python
@pytest.fixture
def temp_file(tmp_path):
    rnd = uuid.uuid4().hex[:8]
    return (tmp_path / f"testfile_{rnd}").name
```

tests/io/test_io.py
```diff
@@ -739,12 +738,12 @@ def test_open_iris1_dataset(iris1_file):
 @pytest.mark.parametrize(
     "compression, compression_opts", [("gzip", 0), ("gzip", 6), ("gzip", 9)]
 )
-def test_odim_roundtrip(odim_file2, compression, compression_opts):
+def test_odim_roundtrip(odim_file2, make_temp_file, compression, compression_opts):
     dtree = open_odim_datatree(odim_file2)
-    with tempfile.NamedTemporaryFile(mode="w+b") as outfile:
+    with open(make_temp_file(), "w+b") as outfile:
```
kmuehlbauer, line 743: Here we could use the second fixture like that:

```python
def test_odim_roundtrip(odim_file2, temp_file, compression, compression_opts):
    dtree = open_odim_datatree(odim_file2)
    xradar.io.to_odim(
        dtree,
        temp_file,
        source="WMO:01104,NOD:norst",
        compression=compression,
        compression_opts=compression_opts,
    )
    dtree2 = open_odim_datatree(temp_file, reindex_angle=False)
    for d0, d1 in zip(dtree.groups, dtree2.groups):
        xr.testing.assert_equal(dtree[d0].ds, dtree2[d1].ds)
```

tests/io/test_io.py
```diff
@@ -754,16 +753,16 @@ def test_odim_roundtrip(odim_file2, compression, compression_opts):
             xr.testing.assert_equal(dtree[d0].ds, dtree2[d1].ds)
 
 
-def test_odim_optional_how(odim_file2):
+def test_odim_optional_how(odim_file2, make_temp_file):
```
kmuehlbauer, line 756: Same here with the second fixture.

tests/io/test_io.py
```diff
@@ -797,22 +796,22 @@ def test_odim_optional_how(odim_file2):
             assert "stopelA" not in ds_how
 
 
-def test_write_odim_source(rainbow_file2):
+def test_write_odim_source(rainbow_file2, make_temp_file):
```
kmuehlbauer, line 799: And here, too.

tests/conftest.py
```diff
@@ -160,3 +161,12 @@ def metek_raw_gz_file(tmp_path_factory):
...
+def make_temp_file(tmp_path):
+    def _make_file():
+        rnd = uuid.uuid4().hex[:8]
+        return tmp_path / f"testfile_{rnd}"
```
kmuehlbauer, line 170: I'd just give back the name here:

```python
return (tmp_path / f"testfile_{rnd}").name
```

tests/io/test_io.py
```diff
@@ -911,15 +910,15 @@ def test_open_datamet_datatree(datamet_file):
 
 
 @pytest.mark.parametrize("first_dim", ["time", "auto"])
-def test_cfradfial2_roundtrip(cfradial1_file, first_dim):
+def test_cfradfial2_roundtrip(cfradial1_file, make_temp_file, first_dim):
```
kmuehlbauer, line 913: Here we can use the first fixture, but we would need to create two filenames and then use the contextmanager like this:

```python
outfile = make_temp_file()
outfile1 = make_temp_file()

xradar.io.to_cfradial2(dtree0.copy(), outfile)
    # then open cfradial2 file
    with xr.open_datatree(outfile) as dtree1:
        # and write again
        fname2 = make_temp_file()
        xradar.io.to_cfradial2(dtree1.copy(), outfile1)
        # and open second cfradial2
        with xr.open_datatree(outfile1) as dtree2:
            # check equality
            for d0, d1, d2 in zip(dtree0.groups, dtree1.groups, dtree2.groups):
               ...
```

kmuehlbauer (changes requested):
Thanks @egouden, I've added a couple comments and suggestion. Hope these make sense to you.

tests/conftest.py
egouden: done

tests/io/test_io.py
egouden: Using make_temp_file is needed ehere

tests/io/test_io.py
egouden: done

tests/conftest.py
egouden: yes

tests/io/test_io.py
egouden: OK

egouden:
I just found a new issue. Using the filename instead of the Path object creates the temporary file in the current working directory and the file is not deleted. It seems that to benefit from the tmp_path fixture the Path must be used. I made the change accordingly.

egouden:
> All green. Ready to merge from your side, @egouden?

Yes

PR #283, opened 2025-07-08
Fix Makefile install command

ReallyNiceGuy:
When calling:
  make install

Use the correct command, as per documentation.


kmuehlbauer:
Thanks Marco!

ReallyNiceGuy:
Sure!

Issue #93, opened 2023-02-27
Implement open_cfradial2_datatree


kmuehlbauer:
* xradar version: 0.1.0
* Python version: 3.11
* Operating System: Linux

### Description

We currently rely on `datatree.open_datatree` for reading CfRadial2 data. Implementing a `open_cfradial2_datatree` aligning with the other readers would help to at least run this through some checks, that the data conforms to the model.

### What I Did

I read a supposed to be CfRadial2 conformant file, but failed later when writing to `ODIM_H5`.

PR #288, opened 2025-07-13
ADD: Cfradial2 Reader

### LINKED ISSUE #287: can xradar read Cfradial2 netcdf files?
Hi all. I just got a file in "cfradial2" format. It contains groups instead a single netcdf. Do we need to create a new backend for this kind of files or we can addapt cfradial1 backend. Attached, you will find a copy of the file

```python 
import xradar as xd

path = "radar-cfradial2.nc"
radar = xd.io.open_cfradial1_datatree(path)
```
returns this error
```python 
Traceback (most recent call last):
  File "/media/alfonso/drive/Alfonso/python/demos/era5-ingestion/truweather.py", line 23, in <module>
    radar = xd.io.open_cfradial1_datatree(path)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alfonso/mambaforge/envs/raw2zarr-dev/lib/python3.12/site-packages/xradar/io/backends/cfradial1.py", line 355, in open_cfradial1_datatree
    "/": _get_required_root_dataset(ds, optional=optional),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alfonso/mambaforge/envs/raw2zarr-dev/lib/python3.12/site-packages/xradar/io/backends/cfradial1.py", line 82, in _get_required_root_dataset
    root = root.rename_vars(
           ^^^^^^^^^^^^^^^^^
  File "/home/alfonso/mambaforge/envs/raw2zarr-dev/lib/python3.12/site-packages/xarray/core/dataset.py", line 4234, in rename_vars
    raise ValueError(
ValueError: cannot rename 'sweep_number' because it is not a variable or coordinate in this dataset
```

this is because is a Netcdf file with nested groups. we can check it out using `xarray.open_groups` or `xarray.open_datatree`

```python 
groups = xr.open_groups(path)

print(groups)

{'/': <xarray.Dataset> Size: 40B
Dimensions:           (nsweep: 1)
Dimensions without coordinates: nsweep
Data variables:
    sweep_group_name  (nsweep) <U10 40B ...
Attributes:
    Conventions:  CF-1.7, Cf/Radial
    version:      2.0
    title:        Radar CfRadial2 Dataset
    history:      Updated CfRadial2 on 2025-05-01T00:46:09.040110Z
    RadarName:    denton.tx
    Latitude:     33.25355
    Longitude:    -97.15204
    Height:       224.0, 

'/sweep_0000': <xarray.Dataset> Size: 27MB
Dimensions:                       (ntime: 723, nrange: 1003)
Dimensions without coordinates: ntime, nrange
Data variables: (12/19)
    time_us                       (ntime) datetime64[ns] 6kB ...
    range_m                       (nrange) float64 8kB ...
    azimuth_deg                   (ntime) float64 6kB ...
    elevation_deg                 (ntime) float64 6kB ...
    range_gate_width_mm           (ntime) float64 6kB ...
    start_range_mm                (ntime) float64 6kB ...
    ...                            ...
    specific_differential_phase   (ntime, nrange) float32 3MB ...
    radial_velocity_mps           (ntime, nrange) float32 3MB ...
    signal_to_noise_ratio_db      (ntime, nrange) float32 3MB ...
    spectral_width_mps            (ntime, nrange) float32 3MB ...
    filtering                     (ntime, nrange) int8 725kB ...
    rain_rate_mmhr                (ntime, nrange) float32 3MB ...
Attributes:
    NumGates:          1003
    ScanType:          2
    AntennaGain:       33.0
    AntennaBeamwidth:  2.7
    PosVelocity:       15.0}
```

Let me know your thoughts.

File can be found in this [link](https://drive.google.com/file/d/1IX5iZnXppTACEmD86vwiihN9tFnKCxvR/view?usp=sharing)

syedhamidali:
- [x] Closes #287, #93
- [x] Tests added


kmuehlbauer (changes requested):
@syedhamidali Just some minor suggestions, but looks good to me. We can enhance the compatibility layer later on, when users request it. Thanks!

tests/io/test_cfradial2.py
```diff
@@ -0,0 +1,41 @@
...
+import tempfile
+from pathlib import Path
+
+import xarray as xr
+from open_radar_data import DATASETS
+
+import xradar as xd
+
+
+def test_open_cfradial2():
```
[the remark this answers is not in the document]
syedhamidali: Done

tests/io/test_cfradial2.py
```diff
@@ -0,0 +1,37 @@
...
+# Distributed under the MIT License. See LICENSE for more info.
+
+import pytest
+import xarray as xr
+from open_radar_data import DATASETS
+
+import xradar as xd
+
+
+@pytest.fixture
```
kmuehlbauer, line 12: @syedhamidali Maybe my suggestion was a bit brief, but this fixture is already defined in conftest.py as well as the tmp_file fixture. You should be able to just use them here and can remove the re-definitions here.

tests/io/test_cfradial2.py
```diff
@@ -0,0 +1,37 @@
...
+def temp_file(tmp_path):
+    return tmp_path / "cfradial2.nc"
+
+
+def test_open_cfradial2(temp_file, cfradial1_file):
+    # Open the original file using CfRadial1 reader
+    dtree = xd.io.open_cfradial1_datatree(cfradial1_file)
+
+    # Write to CfRadial2 NetCDF format
+    dtree.to_netcdf(temp_file)
```
kmuehlbauer, line 27: I'm not really sure, but you might have to add kwarg `engine="netcdf4"` here, since the tmp_file doesn't have an extension.

aladinor (changes requested):
Thanks for putting this together @syedhamidali! Having a CfRadial2 reader is important for xradar. Here are some observations after reviewing the code, tests, and discussion:

## Design

The wrapper approach (`open_datatree` → `open_cfradial2_datatree`) follows KISS and gives API consistency with the other backends. That said, the core concern raised by @kmuehlbauer and myself in the earlier discussion is still open: **CfRadial2 files from different institutions have widely varying compliance with FM301/CfRadial2 spec**.

Without any validation or normalization, users will get inconsistent results:
- Dimension names vary (`ntime`/`nrange` vs `azimuth`/`range`)
- Variable names vary (`azimuth_deg` vs `azimuth`)
- Sweep naming varies (`sweep_0000` vs `sweep_0`)
- Required root attributes may be missing

I think we should either:
- (a) Add a minimal compatibility layer (as @kmuehlbauer suggested), or
- (b) At minimum, add a clear warning in the docstring and/or emit a `UserWarning` that no FM301 validation is performed

## Code

1. **Missing type hints** on the function signature:
   ```python
   # Current
   def open_cfradial2_datatree(filename, **kwargs):
   
   # Suggested
   def open_cfradial2_datatree(filename: str, **kwargs) -> DataTree:
   ```

2. **`zarr` added to `requirements.txt`** — the implementation only uses `xarray.open_datatree` for NetCDF. Why is `zarr` a hard dependency here? If it's needed for a specific use case it should be optional.

## Tests

1. **Fixture re-definitions** — `cfradial1_file` and `temp_file` are re-defined in the test file but already exist in `conftest.py`. @kmuehlbauer flagged this earlier. They should be removed and the conftest fixtures used instead.

2. **Only 1 test (round-trip)** — writing CfRadial1 then reading as CfRadial2. This verifies basic functionality but is missing:
   - Test with kwargs forwarding (e.g., `engine` parameter)
   - Error handling (invalid file path)
   - Verification of actual data values, not just structure (`assert_isomorphic` checks tree shape but not content)

3. **Round-trip coupling** — the test assumes `dtree.to_netcdf()` produces valid CfRadial2. If the export ever breaks, this test fails for the wrong reason. An actual CfRadial2 test file would be more robust.

## Summary

The implementation is clean and minimal, which is good. The main blockers are:
- Remove fixture re-definitions (use conftest.py)
- Justify or remove `zarr` from requirements.txt
- Address the FM301 compliance question (compatibility layer or explicit warning)
- Strengthen tests

Issue #103, opened 2023-04-03
Rainbow reader: _FillValue and _Undetect


egouden:
### Description

In the Rainbow reader, _FillValue is set to 0 and _Undetect is not set. This is causing a loss of information.

Normally _Fillvalue should be set to 255 (or the max value for the byte encoding) and _Undetect to 0.

WMO CF Extensions:

- [WMO306_volume_I4_draft.doc.pdf](https://wmoomm.sharepoint.com/sites/wmocpdb/eve_activityarea/WMO%20Information%20System%20(WIS)_6cf41395-0ffc-ea11-a813-000d3aafe55d/docs%20shared%20on%20community%20site/WMO306_volume_I4_draft.doc.pdf)
- [FM+301-XX+WMO-CF+RADIAL.doc.pdf](https://wmoomm.sharepoint.com/sites/wmocpdb/eve_activityarea/WMO%20Information%20System%20(WIS)_6cf41395-0ffc-ea11-a813-000d3aafe55d/docs%20shared%20on%20community%20site/FM+301-XX+WMO-CF+RADIAL.doc.pdf)

mgrover1:
Thanks for letting us know! Contributions are always welcome :)

kmuehlbauer:
Sorry for the late response, that went somehow under the radar.

Rainbow5 (at least Release 5.57.0) does only have a notion of "no data" which is best translated to `_FillValue` (it's zero for all radar moments on disk). There is no `_Undetect` comparable value in Rainbow5.

For instance a one byte reflectivity looks like this:

0x00 (0) no data 
0x01 (1) -31.5 
0x02 (2) -31.0 
0x40 (64) 0.0 
0x80 (128) 32.0 
0xC0 (192) 64.0 
0xFE (254) 95.0 
0xFF (255) 95.5

We can't do much about the source format.

kmuehlbauer:
And I'm not considering that a bug, as specifying the `_FillValue` is sufficient for the CfRadial2 standard. ODIM_H5 has `nodata` as well as `undetect` mentioned, but if I interpret the documentation correctly both are not mandatory values.

mgrover1:
Ahh okay! Sorry for the mislabelling here @kmuehlbauer , I appreciate the insight!

kmuehlbauer:
No worries @mgrover1, we should invent a label which can be put there instead. How about `data-format-issue`?

mgrover1:
how about `data-standards`?

[...]

PR #290, opened 2025-07-17
Disable fill value in rainbow reader

egouden:
This is the proposed solution for #103.

The "fill value" encoding attribute is set to 0. This introduces NaNs when decoding the values. The origin of these NaNs can be confusing. This is also not consistent with other readers.

The suggestion is to set "fill value" to None. This will not introduce NaNs. This will prevent writing the attribute in the NETCDF file. The default fill value of 255 (uint8) will not be used (this is a valid value in Rainbow format).

This strategy gives room for improved handling of special values. For example using the recommended "flag_values" in FM301.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -1,7 +1,11 @@
 # History
 
-## 0.10.0 (2025-07-11)
+## Development Version (unreleased)
+
 * ENH: Disable fill value in rainbow reader ({issue}`103`) ({pull}`290`) by [@egouden](https://github.com/egouden)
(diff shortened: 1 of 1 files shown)


kmuehlbauer (approved):

egouden:
> @egouden Please add an entry to history.md for visibility. Thanks!

Done.

The entry is a copy of the commit message. Both follow some rules (e.g. references to PR and issue). Could we automate the process?

PR #291, opened 2025-08-01
add nyquist_velocity as coordinate in ODIM H5/GAMIC reader

katelbach:
adds nyquist velocity as a coordinate when using the ODIM H5 and GAMIC backend

- [x] Tests added

### DIFF
--- a/xradar/io/backends/gamic.py
+++ b/xradar/io/backends/gamic.py
@@ -246,12 +246,10 @@ class _GamicH5NetCDFMetadata(_H5NetCDFMetadata):
     @property
     def _nyquist_velocity(self):
         """Return nyquist velocity."""
-        nyquist_vel = self.how.get("nyquist_velocity", None)
-        if nyquist_vel is not None:
-            return float(nyquist_vel)
-        else:
-            pass
-        return None
+        try:
+            return float(self.how["nyquist_velocity"])
+        except (AttributeError, KeyError, TypeError):
+            return None
 
 
 class GamicStore(AbstractDataStore):
--- a/xradar/io/backends/gamic.py
+++ b/xradar/io/backends/gamic.py
@@ -243,6 +243,16 @@ class _GamicH5NetCDFMetadata(_H5NetCDFMetadata):
         """Return sweep number."""
         return int(self._group.split("/")[0][4:])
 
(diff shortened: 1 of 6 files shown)


xradar/io/backends/odim.py
```diff
@@ -477,6 +486,17 @@ def _sweep_number(self):
...
+    @property
+    def _nyquist_velocity(self):
+        try:
+            nyquist_vel = self.how.get("NI", None)
+            if nyquist_vel is not None:
+                return float(nyquist_vel)
+        except (AttributeError, KeyError, TypeError):
+            pass
+
+        return None
```
kmuehlbauer, line 498: We can simplify this a bit, if we already catch the possible errors. If there are issues that "NI" might contain something which breaks the cast to float, we would need to add `ValueError` to that list.

```suggestion
        try:
            return float(self.how["NI"])
        except (AttributeError, KeyError, TypeError):
            return None
```

xradar/io/backends/gamic.py
```diff
@@ -243,6 +243,16 @@ def _sweep_number(self):
...
 
+    @property
+    def _nyquist_velocity(self):
+        """Return nyquist velocity."""
+        nyquist_vel = self.how.get("nyquist_velocity", None)
+        if nyquist_vel is not None:
+            return float(nyquist_vel)
+        else:
+            pass
+        return None
```
kmuehlbauer, line 254: We can simplify this too, according to the odim example.

xradar/model.py
```diff
@@ -727,6 +728,14 @@ def get_time_attrs(date_str="1970-01-01T00:00:00Z", date_unit="seconds"):
     return time_attrs
 
 
+def get_nyquist_velocity_attrs(units="meters per second"):
```
kmuehlbauer, line 731:
```suggestion
def get_nyquist_velocity_attrs(units="m s-1"):
```

kmuehlbauer (approved):
@katelbach This is ready for merge, I think! Please add an entry to `history.md`

katelbach:
done, thanks for your guiding me through the process @kmuehlbauer

kmuehlbauer (approved):

Issue #243, opened 2024-11-17
ENH: Consider adding lat, lon and alt option in georeferencing


syedhamidali:
* xradar version: 0.8.1
* Python version: 3.12.7
* Operating System: MacOS Sequoia 15.1 (24B83)

### Description
The current georeferencing in xradar provides Cartesian coordinates (x, y, z), but there’s no direct option to include geographic coordinates (lat, lon, and alt). Adding this feature would enhance usability for users needing georeferenced datasets in geographic space.

#### Attempted to use georeferencing with geo=True

```python
geo = radar.xradar.georeference(geo=True)
```

kmuehlbauer:
@syedhamidali Thanks for this proposal. 

In https://github.com/openradar/xradar/issues/25#issuecomment-1255448361 I've proposed to just implement radar coordinates as `AEQD` coordinates, and leave projecting/transforming coordinates to specialized libraries (eg. `Py-ART`/`wradlib` or even more projection dedicated ones like `pyproj`).

During enhancing plotting capabilities we've finally implemented a way to use the `CRS` coordinate (eg. `spatial_ref`, `crs_wkt`)  via `pyproj` to have cartopy do the transformation.   

There is also work still going on to build this `CRS` as Index accessor-based into xarray (https://github.com/pydata/xarray/issues/2288), but it's hard to predict, when this will be in a usable state (need to re-check!)

From my perspective we should not implement the conversion/transformation into xradar as proposed in #244, but guide users how to do this with these external libraries (which they might use anyway in further processing). When using our new `map_over_sweeps` accessor this is only one or two additional lines of code. 

If we (ping @openradar/developers) still want to have this functionality available from `.xradar.georeference()` I'd suggest a slightly different API:

```python
ds = ds.xradar.georeference(crs=trg_crs, names={"x": "lon", "y": "lat", "z": "alt"})
```

Internally we would setup `pyproj.Transformer` and do the transform from xradars AEQD to any possible projection/coord system.

Here is the functionality as MCVE:

```python
import numpy
import xarray as xr
import xradar
from pyproj import CRS, Transformer, transform
from open_radar_data import DATASETS

filename = DATASETS.fetch("cfrad.20080604_002217_000_SPOL_v36_SUR.nc")
radar = xd.io.open_cfradial1_datatree(filename, first_dim="auto")

swp = radar["sweep_0"].to_dataset().xradar.georeference()

src_crs = swp.xradar.get_crs()
[...]

PR #300, opened 2025-08-26
add option to reproject radar coordinates

egouden:
Reprojection is needed by many applications but not easy for the average user.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -10,6 +10,7 @@
 * FIX: Return float32 for float types (keep other dtypes) and remove erroneous "units" from time attrs for hpl reader. ({issue}`296`) ({pull}`297`) by [@kmuehlbauer](https://github.com/kmuehlbauer)
 * MNT: Fix Windows errors in file handling ({pull}`295`) by [@egouden](https://github.com/egouden)
 * MNT: do not overwrite Dataset for Dataset.update ({pull}`302`) by [@kmuehlbauer](https://github.com/kmuehlbauer)
+* FIX: Properly handle zero and NaN values during IRIS/Sigmet KDP decode. ({pull}`301`) by [@billvieux](https://github.com/billvieux)
 
 ## 0.10.0 (2025-07-11)
 
--- a/tests/io/test_iris.py
+++ b/tests/io/test_iris.py
@@ -90,19 +90,24 @@ def test_decode_velc():
 
 def test_decode_kdp():
     np.testing.assert_array_almost_equal(
-        iris.decode_kdp(np.arange(-5, 5, dtype="int8"), wavelength=10.0),
-        [
-            12.243229,
-            12.880858,
-            13.551695,
-            14.257469,
-            15.0,
-            -0.0,
-            -15.0,
-            -14.257469,
-            -13.551695,
-            -12.880858,
-        ],
+        iris.decode_kdp(
+            np.array(
+                [
+                    0,
+                    1,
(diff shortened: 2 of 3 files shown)


docs/history.md
```diff
@@ -2,6 +2,7 @@
 
 ## Development Version (unreleased)
 
+* ENH: add option to reproject radar coordinates
```
kmuehlbauer, line 5: Please add issue, pr and your credentials

xradar/georeference/projection.py
```diff
@@ -133,3 +128,25 @@ def add_crs_tree(radar, datum="WGS84"):
         if "sweep" in key:
             radar[key].ds = add_crs(radar[key].to_dataset(), datum=datum)
     return radar
+
+
+def radar_crs(ds):
```
kmuehlbauer, line 133: Should we make this available also as accessor?

`ds.xradar.radar_crs`

xradar/georeference/transforms.py
```diff
@@ -149,20 +153,26 @@ def get_x_y_z(ds, earth_radius=None, effective_radius_fraction=None):
...
-    ds["x"], ds["y"], ds["z"] = antenna_to_cartesian(
+    x, y, z = antenna_to_cartesian(
         ds.range,
         ds.azimuth,
         ds.elevation,
         earth_radius=earth_radius,
         effective_radius_fraction=effective_radius_fraction,
         site_altitude=ds.altitude.values,
     )
+    ds["x"] = x
```
kmuehlbauer, line 173: If we do it like this, maybe add the attributes from below here, too.

kmuehlbauer (changes requested):
@egouden These additions will improve handling with projections. I've added a couple comments. This will also need tests. Sorry for the delay in review.

kmuehlbauer:
@egouden Could you add some example how to use this? Please also have a look at #243 for some discussion on that.

xradar/georeference/transforms.py
egouden: I improved the handling.

kmuehlbauer:
Thanks for the example, @egouden.

Would it make sense to make this available like this, then? Can be done in follow-up PR after discussion on the wanted API.

```python
radar = radar.xradar.georeference(trg_crs=4326)
```

xradar/georeference/transforms.py
```diff
@@ -149,28 +152,72 @@ def get_x_y_z(ds, earth_radius=None, effective_radius_fraction=None):
...
+    # Transform coordinates to target_crs if provided
+    if target_crs is not None:
+
+        target_crs = pyproj.CRS(target_crs)
+
+        transformer = pyproj.Transformer.from_crs(crs, target_crs, always_xy=True)
+        x_new, y_new = transformer.transform(ds.x.values, ds.y.values)
+
+        ds["x"] = (("azimuth", "range"), x_new.reshape(ds.x.shape))
+        ds["y"] = (("azimuth", "range"), y_new.reshape(ds.y.shape))
```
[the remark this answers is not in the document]
egouden: This is a difficult problem. I would keep z as amsl (https://epsg.io/5773) for now.

kmuehlbauer (approved):
Thanks @egouden!

kmuehlbauer:
Ok, let's get this in. Thanks @egouden

kmuehlbauer (approved):

PR #301, opened 2025-09-14
Fix IRIS/Sigmet KDP decode

billvieux:
Use correct mask for zero values. Use np.nan for unavailable values. Improve test coverage of special values.

### DIFF
--- a/xradar/io/backends/iris.py
+++ b/xradar/io/backends/iris.py
@@ -218,7 +218,10 @@ def decode_kdp(data, **kwargs):
     See 4.4.20 p.77
     """
     wavelength = kwargs.pop("wavelength")
-    zero = data[data == -128]
+    zero = data == -128
+    not_available = (data == 0) | (data == -1)
+    data = data.astype(np.float64)
+    data[not_available] = np.nan
     data = -0.25 * np.sign(data) * 600 ** ((127 - np.abs(data)) / 126.0)
     data /= wavelength
     data[zero] = 0
--- a/docs/history.md
+++ b/docs/history.md
@@ -9,6 +9,7 @@
 * FIX: Return float32 for float types (keep other dtypes) and remove erroneous "units" from time attrs for hpl reader. ({issue}`296`) ({pull}`297`) by [@kmuehlbauer](https://github.com/kmuehlbauer)
 * MNT: Fix Windows errors in file handling ({pull}`295`) by [@egouden](https://github.com/egouden)
 * MNT: do not overwrite Dataset for Dataset.update ({pull}`302`) by [@kmuehlbauer](https://github.com/kmuehlbauer)
+* FIX: Properly handle zero and NaN values during IRIS/Sigmet KDP decode. ({pull}`301`) by [@billvieux](https://github.com/billvieux)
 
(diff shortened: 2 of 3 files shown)


xradar/io/backends/iris.py
```diff
@@ -218,7 +218,10 @@ def decode_kdp(data, **kwargs):
     See 4.4.20 p.77
     """
     wavelength = kwargs.pop("wavelength")
-    zero = data[data == -128]
+    zero = data == -128
```
kmuehlbauer, line 221: @egouden You mentioned the difference between "no data available" and "area not scanned". We might not have good solution to this, but would it make sense here to bind those two together and set to nan?

billvieux:
Thanks @egouden! The calculation for non-special values was already correct as far as I can tell. The implementation is a bit unusual and does not reflect the two equations in the manual directly, but mathematically it results in the same value. I intentionally did not try to rewrite the calculation for this fix. 

The issues with the current decoding are
1. data[data == -128] is incorrectly used as a mask rather than (data == -128). The wrong values in the output are set to zero.
2. The not available and not scanned special values are decoded as valid KDP values and returned. 
3. The test does not include the not available and not scanned special values. 
(Note: There is a small typo in the documentation table with example KDP values for uint 254 at 10cm which should be same as 2 at 10cm except positive. Other than that, the test now reflects the documentation's example values.)

billvieux:
One other clarification - the current decoder processes signed int (int8) values. The documentation describes unsigned int (uint8). I included the uint8 in the test comments for clarity when comparing with documentation. The -128 the code's zero mask is a signed int and the same as the documentation's 128 unsigned int.

egouden:
Looks good now with the proper mapping ;-)

egouden:
Thanks @billvieux for the fix !

Have you verified that the other variables have been decoded correctly? Getting the correct values is critical for the readers. I suggest to add plots of all variables in the examples to properly verify this.

It seems the IRIS data are coded in uint8 but stored as int8 for historical reasons. I suggest to convert the data to uint8 directly before decoding to improve readability and avoid mistakes. Accordingly in the test the uint8 values to be tested can be converted to int8.

billvieux:
I have not reviewed the decoding of the other variables. The reference documentation refers to both unsigned and signed values throughout (including their "Scalar Definitions" section). As @kmuehlbauer mentions, the KDP field documentation says it is stored as signed, but then provides a table of examples using the unsigned representation. And that table has a typo in it for the 254 / 10 cm value. So the documentation is not the most consistent. :)

kmuehlbauer:
I've digged in the old wradlib code, when I first implemented this functionality. I do not really remember all the details, but my main concern was to have only one decoding function for the whole array.

The function as it is now will decode both parts (positive and negative) in one go. This is not possible when using unsigned int. That way we would have to declare two functions and work on values < 128 and values > 128 separately. 

I'd be in favour of adding more comments to the code, describing the behaviour in more detail.

For the decoding of the variables we have according test functions. These test functions should test correct behaviour. We still might have bugs, because of slightly incorrect decoding/test functions. So I'm also in support of @egouden's idea to add these evaluation plots.

PR #315, opened 2025-12-09
enhance contributing experience

egouden:
* update guide
* update requirements

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,8 +3,6 @@
 Contributions are welcome, and they are greatly appreciated! Every little bit
 helps, and credit will always be given.
 
-You can contribute in many ways:
-
 ## Types of Contributions
 
 ### Report Bugs
@@ -40,7 +38,7 @@ The best way to send feedback is to file an issue at [xradar-issues](https://git
 If you are proposing a feature:
 
 * Explain in detail how it would work.
-* Keep the scope as narrow as possible, to make it easier to implement.
+* Keep the scope as narrow as possible to make it easier to implement.
 * Remember that this is a volunteer-driven project, and that contributions
   are welcome :)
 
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
> There are some files not deleted after notebook tests. @kmuehlbauer Is there a quick fix?


Need to look this up. We might be able to get away with this, if we run the tests in a dedicated test folder, which will be cleared afterwards.

CONTRIBUTING.md
```diff
@@ -49,94 +47,101 @@ If you are proposing a feature:
...
+        python -m sphinx build -j auto -v -b html docs/ doc-build
+        ```
+    - Check at `docs/_build/html/index.html`
+    - Add new features to README.rst
+
+7. Run pre-commit hooks
+
+    ```bash
+    git add your-new-file.py
+    pre-commit run --all-files
```
kmuehlbauer: Do we need to tell the user to do `pre-commit install` once before running?

kmuehlbauer (commented):
Looks good, but will require additional maintenance. We should take that path, it will make life easier for users new to the field.

Only one question regarding pre-commit.

CONTRIBUTING.md
egouden: I would let the developer decides.

egouden: The guide should be concise. Let's recommend pre-commit install by default !

egouden:
> > There are some files not deleted after notebook tests. @kmuehlbauer Is there a quick fix?
> 
> Need to look this up. We might be able to get away with this, if we run the tests in a dedicated test folder, which will be cleared afterwards.

I think this fix should be integrated in the PR.

kmuehlbauer:
> > > There are some files not deleted after notebook tests. @kmuehlbauer Is there a quick fix?
> > 
> > 
> > Need to look this up. We might be able to get away with this, if we run the tests in a dedicated test folder, which will be cleared afterwards.
> 
> I think this fix should be integrated in the PR.

See #316 for a fix.

kmuehlbauer (approved):

PR #324, opened 2026-02-03
use dtype with fallback

katelbach:
Minor fix as I noted that ODIM H5 exports fail because the fallback value for dtype is not used in an if query.

- [ ] Tests added


kmuehlbauer (approved):
Thanks @katelbach! I think we can get this now. All in place!

PR #337, opened 2026-02-27
ENH: skip redundant station coord reads for subsequent sweeps in DataTree

### LINKED ISSUE #334: ENH: skip redundant station coord reads for subsequent sweeps in DataTree
## Summary

Follow-up optimization from #333 (suggested by @kmuehlbauer in [PR review](https://github.com/openradar/xradar/pull/333#discussion_r2834710219)).

When building a DataTree, every sweep currently reads `latitude`, `longitude`, and `altitude` from the store, even though the values are identical across all sweeps. Since `_assign_root()` already extracts station coords from sweep_0 and places them as coordinates on the root node, and DataTree coordinate inheritance passes them down to child sweep nodes, the reads for sweeps 1+ are redundant I/O.

## Proposed change

Use `site_coords=True` only for the first sweep and `site_coords=False` for all subsequent sweeps when constructing a DataTree. This slightly shifts the meaning of `site_coords` — it would control whether station coords are read from the store, not just whether they appear as coordinates.

This applies to all backends that call `open_*_dataset()` per sweep inside their `open_*_datatree()` functions.

## Impact

- Negligible for a single volume
- Adds up significantly when processing hundreds or thousands of files (e.g., multi-day radar archives)
- Reduces I/O by `2 * (n_sweeps - 1)` scalar reads per volume (3 variables x (n_sweeps - 1) sweeps)

## Related

- #331 — original issue for station coords on root
- #333 — PR that promoted station coords to root coordinates

aladinor:
## Summary

Move station coordinates (`latitude`, `longitude`, `altitude`) to the root node as coordinates and **drop them entirely from sweep nodes**. This ensures the root is the single authoritative source for station location in a DataTree.

- Rename boolean kwarg from `site_coords` to `site_as_coords` for clarity (per review)
- Extract shared `_apply_site_as_coords()` helper into `common.py`, replacing duplicated if/else blocks across all backends
- Refactor `_assign_root()` to return `(root, cleaned_sweeps)` — no hidden side effect
- Drop station vars in `_attach_sweep_groups()` for backends using that path
- Fix read-only array bug in `_ipol_time` using `da.where` instead of in-place mutation

Closes #334

## Dependency

> **Do not merge until [pydata/xarray#11230](https://github.com/pydata/xarray/pull/11230) is merged.**

Xarray's `DataTree` currently does not inherit non-index coordinates from parent nodes. The xarray PR adds `inherit="all"` support, which will allow sweeps to access `latitude`, `longitude`, and `altitude` from the root via `to_dataset(inherit="all")`.

## Follow-up PR (after xarray PR merges)

Update the georeferencing module to use `to_dataset(inherit="all")` so `ds.latitude`, `ds.longitude`, `ds.altitude` are accessible on sweep datasets:

- `xradar/georeference/projection.py:129` — uses `ds.latitude.values` and `ds.longitude.values`
- `xradar/georeference/transforms.py:255` — uses `ds.latitude.values` and `ds.altitude.values`

## Changes

- `xradar/io/backends/common.py`: Add `_apply_site_as_coords()` helper; refactor `_assign_root()` to return cleaned sweeps; drop station vars in `_attach_sweep_groups()`
- All backend `open_dataset` methods: Use `_apply_site_as_coords(ds, site_as_coords)`
- All `open_*_datatree()` functions: Pass `site_as_coords=False` for sweep datasets
- `xradar/io/backends/cfradial1.py`: Always assign station vars to sweeps before calling `_apply_site_as_coords`
- `xradar/io/backends/nexrad_level2.py`, `xradar/io/backends/uf.py`: Use `zip()` to build sweep entries from cleaned `ls_ds`
- `xradar/util.py`: Fix read-only array in `_ipol_time` with `da.where`
- `docs/history.md`: Add changelog entry

## Test plan

- [x] All 440 tests pass
- [x] `TestStationCoordsOnRoot` verifies root has station coords as coordinates and sweeps have none (ODIM, NEXRAD, GAMIC, CfRadial1)
- [x] Test data_vars counts updated to reflect 3 fewer vars per sweep
- [x] `ruff check` and `black --check` pass clean


xradar/io/backends/common.py
```diff
@@ -63,6 +63,37 @@ def _fix_angle(da):
 _STATION_VARS = {"latitude", "longitude", "altitude"}
 
 
+def _apply_site_coords(ds, site_coords):
+    """Promote or demote station coordinates on a sweep Dataset.
+
+    When *site_coords* is true the latitude / longitude / altitude
```
kmuehlbauer: @aladinor Thanks, this is looking good to me. Just one thing, would it make sense to use ``site_as_coords``? This would be a bit more explicit. WDYT?

xradar/io/backends/iris.py
```diff
@@ -3891,7 +3892,7 @@ def open_store_coordinates(self, var):
         prt_mode = "not_set"
         follow_mode = "not_set"
 
-        lon, lat, alt = self.root.site_coords
+        lon, lat, alt = self.root.site_as_coords
```
kmuehlbauer: Same

kmuehlbauer (changes requested):
There are occasions of `site_coords` replaced which are related to actual `site_coords` tuple, not the boolean kwarg.

Please revert these part.

kmuehlbauer (approved):

aladinor:
yeap. we are ready to go.

PR #344, opened 2026-03-27
FIX: allow passing inherit parameter to apply_to_sweeps

### LINKED ISSUE #343: ENH: allow passing inherit parameter to apply_to_sweeps / map_over_sweeps
## Summary

`apply_to_sweeps()` hardcodes `inherit="all_coords"` in the `to_dataset()` call (line 572 of `util.py`). This means every function applied via `map_over_sweeps` or `apply_to_sweeps` always inherits root coordinates (lat/lon/alt) onto sweep datasets — even when the function doesn't need them (e.g. `reindex_angle`).

This causes a side effect: after `map_over_sweeps` runs, the inherited coords become **locally stored** on each sweep node, defeating the purpose of PR #337 which moved station coords to the root only.

## Current behavior

```python
# util.py line 572
func(
    dtree[node.path].to_dataset(inherit="all_coords"),  # hardcoded
    *args,
    **kwargs  # goes to func, not to_dataset
)
```

Calling `dtree.xradar.map_over_sweeps(reindex_angle)` leaks lat/lon/alt onto every sweep, even though `reindex_angle` only needs azimuth/range.

## Proposed fix

Pop `inherit` from `kwargs` in `apply_to_sweeps` before forwarding to `func`:

```python
def apply_to_sweeps(dtree, func, *args, **kwargs):
    inherit = kwargs.pop("inherit", "all_coords")
    ...
    tree.update({
        node.path: func(
            dtree[node.path].to_dataset(inherit=inherit),
            *args,
            **kwargs
        )
        ...
    })
```

This allows callers to control inheritance:

```python
# Functions that don't need lat/lon/alt
dtree.xradar.map_over_sweeps(reindex_angle, inherit=False)

# Functions that need lat/lon/alt (default, backwards compatible)
dtree.xradar.map_over_sweeps(georeference_func)
```

## Context

Discovered while integrating xradar PR #337 into [raw2zarr](https://github.com/aladinor/raw2zarr). After `fix_angle` (which uses `map_over_sweeps`), station coords re-appear on sweeps and must be cleaned up manually.

aladinor:
## Summary

- Pop `inherit` from `kwargs` in `apply_to_sweeps` before forwarding to `func`, defaulting to `"all_coords"` for backward compatibility
- This lets callers control whether root coords (lat/lon/alt) are inherited onto sweep datasets via `to_dataset(inherit=...)`
- Using `inherit=False` prevents station coords from leaking onto sweeps after `map_over_sweeps`, preserving the root-only storage from PR #337

Closes #343

## Test plan

- [x] `test_apply_to_sweeps_inherit_false` — verifies `inherit=False` prevents root coord leakage
- [x] `test_apply_to_sweeps_inherit_via_map_over_sweeps` — verifies it works through the accessor
- [x] Existing `test_apply_to_sweeps` still passes (backward compatible)
- [x] All 130 util + accessor tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -10,6 +10,7 @@
 * ADD: ``incomplete_sweep`` parameter (``"drop"``/``"pad"``) to ``open_nexradlevel2_datatree`` for handling incomplete sweeps in partial volume data ({pull}`332`) by [@aladinor](https://github.com/aladinor)
 * ADD: Notebook example for streaming NEXRAD Level 2 chunks from S3 (``nexrad_read_chunks.ipynb``) ({pull}`332`) by [@aladinor](https://github.com/aladinor)
 * ADD: Comprehensive test suite for chunk list-input and incomplete sweep handling ({pull}`332`) by [@aladinor](https://github.com/aladinor)
+* FIX: Allow passing ``inherit`` parameter to ``apply_to_sweeps`` / ``map_over_sweeps`` to control coordinate inheritance from root node ({issue}`343`, {pull}`344`) by [@aladinor](https://github.com/aladinor)
 
 ## 0.11.1 (2026-02-03)
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,14 @@ def file_or_filelike(request):
 
 
 @pytest.fixture(scope="session")
+def cfradial1_sgp_dtree():
+    import xradar as xd
+
+    filename = DATASETS.fetch("sample_sgp_data.nc")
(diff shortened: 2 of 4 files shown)


syedhamidali (approved):

kmuehlbauer (changes requested):
Two minor things, otherwise LGTM!

kmuehlbauer (approved):
Thanks @aladinor!

PR #345, opened 2026-03-27
FIX: handle file-like

### LINKED ISSUE #189: Open Data using File-Like Object
* xradar version: 0.6.0
* Python version: 3.11.5
* Operating System: Alma Linux

Is it possible to use a file-like object instead of filenames when using xradar? According to the xarray open_dataset documentation, as well as the [xradar Odim Backend Documentation](https://docs.openradarscience.org/projects/xradar/en/stable/generated/xradar.io.backends.odim.OdimBackendEntrypoint.html#xradar.io.backends.odim.OdimBackendEntrypoint), file objects are supported, but with the 'odim' engine, I get an error:
```python
from open_radar_data import DATASETS
import xarray as xr
import xradar as xd
filename = DATASETS.fetch("71_20181220_060628.pvol.h5")
#works:
ds = xr.open_dataset(filename, group="sweep_0", engine="odim") 
#works as well (at least no exception raised):
with open(filename, 'rb') as f:
    ds = xr.open_dataset(f)
# Raises TypeError: cannot pickle '_io.BufferedReader' object
with open(filename, 'rb') as f:
    ds = xr.open_dataset(f, group="sweep_0", engine="odim")
```

Thank you for any help and the xradar development in general!
Paul

kmuehlbauer:
- [x] Closes #189
- [x] Tests added/updated


kmuehlbauer:
Well, that worked! Now checking for other backends.

PR #362, opened 2026-04-20
FIX: open_nexradlevel2_datatree keeps sweeps with interior gaps

### LINKED ISSUE #361: open_nexradlevel2_datatree fails on files with interior sweep-index gap (573 KLOT volumes, 0.68% rate)
## Summary

`open_nexradlevel2_datatree` fails with `KeyError` on NEXRAD Level-II files that are missing one interior sweep — even though **every remaining sweep in the file contains valid data with correct elevation angles**. The cause is in `nexrad_level2.py:2123` (and symmetrically `:2140`): sweep names are generated from `range(act_sweeps)` rather than from the actual keys in `nex.msg_31_data_header`. When NOAA's upstream pipeline drops a single sweep and the remaining sweeps retain their original index labels, xradar requests an index that no longer exists.

This is not a theoretical corner case. Across a 14-month NEXRAD KLOT ingest (~84K volumes) against `s3://unidata-nexrad-level2/` we hit this bug on **573 files (≈0.7% of all volumes)**. Every one of them contains valid radar data for every sweep that is present.

## Reproduction

```python
from xradar.io.backends.nexrad_level2 import NEXRADLevel2File
import xradar as xd
import fsspec

path = "s3://unidata-nexrad-level2/2021/11/04/KLOT/KLOT20211104_174202_V06"

with fsspec.open(path, mode="rb", anon=True) as fh:
    buf = fh.read()

nex = NEXRADLevel2File(buf)
print("MSG_5 declared cuts:", nex.get_msg_5_data()["number_elevation_cuts"])
print("len(msg_31_header):  ", len(nex.msg_31_header))
print("data.keys():         ", sorted(nex.data.keys()))

# → MSG_5 declared cuts: 12
# → len(msg_31_header):   11
# → data.keys():          [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11]

dt = xd.io.open_nexradlevel2_datatree(buf)
# → KeyError: 10
```

xradar version: `0.11.2.dev10+g6cba751ce` (but the affected lines have been present for a while).

## Root cause — `nexrad_level2.py:2123`

```python
# Current
with NEXRADLevel2File(filename_or_obj, loaddata=False) as nex:
    act_sweeps = len(nex.msg_31_data_header)           # 11
    ...

if incomplete_sweep == "drop":
    sweeps = [f"sweep_{i}" for i in range(act_sweeps)  # range(11) = [0..10]
              if i not in incomplete]
    ...
elif incomplete_sweep == "pad":
    sweeps = [f"sweep_{i}" for i in range(act_sweeps)]  # same bug
```

`msg_31_data_header` is an `OrderedDict` (or equivalent) keyed by the file's **actual** sweep indices — which in these affected files are `[0, 1, 2, ..., 9, 11]`, not `[0, 1, ..., 10]`. `range(act_sweeps)` produces `[0..act_sweeps-1]`, which does not match the actual keys when any key is missing in the middle.

Downstream, when the backend tries `self.data[10]` (at `nexrad_level2.py:636` via `get_sweep` / `get_moment`), the key doesn't exist and a `KeyError: 10` bubbles up.

## Proposed fix (one line, same shape in both branches)

```python
# nexrad_level2.py:2122–2145 (cleaned up)
with NEXRADLevel2File(filename_or_obj, loaddata=False) as nex:
    present_keys = sorted(nex.msg_31_data_header.keys())
    incomplete = nex.incomplete_sweeps
    ...

if incomplete_sweep == "drop":
    sweeps = [f"sweep_{i}" for i in present_keys if i not in incomplete]
elif incomplete_sweep == "pad":
    sweeps = [f"sweep_{i}" for i in present_keys]
```

Use the **actual keys** present in `msg_31_data_header`, not a positional `range(len(...))`. Same one-line fix at both the `drop` and `pad` branches.

**`exp_sweeps`-based AVSET logic at line 2119 is also affected** — it compares counts but not keys. It should still work for the AVSET case (where the actual sweeps are fewer than the VCP schedule AND the keys are contiguous `[0..N-1]`), but for the "interior missing sweep" case it would leave a gap that downstream code doesn't expect. Worth double-checking in the PR.

## Evidence the files are not actually corrupted

Every sweep present in each affected file has a valid elevation angle matching the MSG_5 declared schedule for that specific index. Ran the comparison on 3 representative files:

### File 1 — `KLOT20211104_174202_V06` (VCP with 12 declared cuts → 11 present, gap at index 10)

```
 cut     declared       actual        match
 ----    ------------   ------------  -----
   0     0.4833984375   0.50537109375   OK
   1     0.4833984375   0.3955078125    OK
   2     0.87890625     0.7965087890625 OK
   3     0.87890625     0.8349609375    OK
   4     1.318359375    1.31011962890625 OK
   5     1.318359375    1.241455078125  OK
   6     1.8017578125   1.77978515625   OK
   7     2.4169921875   2.2247314453125 OK
   8     3.1201171875   3.00201416015625 OK
   9     3.9990234375   3.8397216796875 OK
  10     5.09765625     MISSING         -   ← single-sweep gap, upstream-dropped
  11     6.416015625    6.339111328125  OK   ← kept its ORIGINAL index
```

### File 2 — `KLOT20211119_162142_V06` (8 declared → 7 present, gap at index 1)

```
   0     0.4833984375   0.3680419921875   OK
   1     0.4833984375   MISSING           -
   2     1.494140625    1.46942138671875  OK
   3     1.494140625    1.4501953125      OK
   4     2.5048828125   2.56256103515625  OK
   5     2.5048828125   2.4609375         OK
   6     3.515625       3.570556640625    OK
   7     4.482421875    4.5098876953125   OK
```

### File 3 — `KLOT20220424_151601_V06` (18 declared → 17 present, gap at index 11)

```
   0     0.48°    actual 0.52°    OK
   1     0.48°    actual 0.44°    OK
   2–10  match declared within 0.2°  OK
  11     6.42°    MISSING          -
  12     8.00°    actual 7.82°    OK
  13–17  match declared            OK
```

In every case: **the data for each present sweep is valid, the elevation angle is correct, and the "missing" index always sits cleanly in the interior** (often at a split-cut boundary or a VCP transition).

## What's actually happening upstream

Based on the pattern across 573 files over 14 months:

1. NOAA's RDA/RPG → Unidata S3 pipeline serializes one fewer sweep than MSG_5 declares (`number_elevation_cuts = 12`, but only 11 sweeps land in the final file).
2. The remaining sweeps **keep their original indices** from the complete N-cut plan (e.g., `[0..9, 11]` instead of `[0..10]`).
3. `len(nex.msg_31_data_header) = 11`, but `max(nex.msg_31_data_header.keys()) = 11`.

Hypothesis is consistent with the "lost split-cut pair" shape in several files (missing index is the second of a split-cut pair), but the bug in xradar manifests regardless of WHY the sweep is missing — any interior gap in the keys trips it.

### Affected files (production data)

`s3://nexrad-arco/KLOT/` ingestion with [`raw2zarr`](https://github.com/aladinor/raw2zarr) (which wraps `xd.io.open_nexradlevel2_datatree`):

| Window | Total volumes | CORRUPTED (this bug) | Rate |
|---|---|---|---|
| Nov 2021 back-fill | 6,607 | 43 | 0.65% |
| 2022 (Jan–Dec) | 77,809 | 530 | 0.68% |
| **Total (14 months, KLOT only)** | **84,416** | **573** | **0.68%** |

Extrapolated across all NEXRAD sites and years, this bug likely drops hundreds of thousands of otherwise-valid volumes on the floor globally.

### Pattern distribution

Aggregating the "extra" sweep index (the one xradar wrongly reports as present beyond its count):

| `extra=[X]` | count | interpretation |
|---|---|---|
| `[11]` | 382 | VCP declares 12 cuts, file has 11 (indices 0–9, 11) |
| `[7]`  | 53  | VCP declares 8, file has 7 |
| `[16]` | 23  | VCP declares 17, file has 16 |
| `[17]` | 20  | VCP declares 18 |
| `[14]` | 13  | |
| `[20]` | 10  | |
| `[15]` | 10  | |
| `[13]` | 7   | |
| `[12]` | 7   | |
| `[10, 11]` | 6  | two sweeps dropped |
| ... | ~ rest | long tail |

In every case the "extra" index is exactly `declared_cuts - 1` from the original VCP plan.

## Suggested test

Add to `xradar/tests/io/test_nexrad.py` a fixture that:

1. Builds a minimal synthetic Level-II buffer where `msg_5["number_elevation_cuts"] = 5` but `msg_31_data_header` has only 4 entries keyed as `{0, 1, 2, 4}` (skipping index 3).
2. Calls `open_nexradlevel2_datatree(buf)` — currently raises `KeyError: 3`; after the fix returns a `DataTree` with `sweep_0, sweep_1, sweep_2, sweep_4`.
3. Verifies each returned sweep has the expected elevation angle.

Alternatively, any of the three S3 paths above can stand in as a black-box regression fixture (they're on public NOAA NODD).

## Context

- Downstream raw2zarr investigation: [aladinor/raw2zarr#210](https://github.com/aladinor/raw2zarr/issues/210)
- Our workaround (while this lands): pre-read the file with `NEXRADLevel2File(..., loaddata=False)`, extract `sorted(nex.msg_31_data_header.keys())`, pass explicitly as `sweep=[...]` to `open_nexradlevel2_datatree`. Works around line 2123 by using the explicit-list branch (line 2109) instead. Restores 573 volumes per 84K in our KLOT corpus.

Happy to open a PR with the fix + a regression test if that helps.

aladinor:
## Summary

- Resolves #361.
- Derive sweep names from the actual sweep indices in `nex.data` (`sorted(nex.data)`) instead of `range(len(nex.msg_31_data_header))`. When NOAA's upstream pipeline drops an interior cut, the surviving sweeps keep their original labels (e.g. `[0..9, 11]` when cut 10 is missing); the old logic asked for the non-existent `sweep_10` and raised `KeyError` downstream, discarding the entire volume even though every present sweep contained valid data.
- Observed on ~0.7% of real NEXRAD Level-II volumes (573/84K across 14 months of KLOT on `s3://unidata-nexrad-level2/`).

## Implementation notes

- One-line change per branch in `open_nexradlevel2_datatree` (`xradar/io/backends/nexrad_level2.py`), applied to both `incomplete_sweep="drop"` and `incomplete_sweep="pad"`.
- `nex.incomplete_sweeps` is now read before `nex.data`, because it implicitly triggers `data_header` parsing which populates `nex.data`. Order matters; a comment documents this.
- AVSET-mode clamp (`if exp_sweeps > act_sweeps: exp_sweeps = act_sweeps`) is unchanged — `act_sweeps` remains the count of present sweeps.

## Test plan

- [x] New parametrized regression test in `TestIncompleteSweepParameter`, 7 cases:
  - single interior gap (drop + pad)
  - gap combined with an incomplete sweep (drop removes both; pad keeps the present one)
  - multiple interior gaps
  - trailing gap (MSG_5 declared N cuts, parser got fewer)
  - unsorted dict-iteration order (guards the `sorted()` call)
- [x] Full `tests/io/test_nexrad_level2.py` passes (104 tests).
- [x] `black --check` and `ruff check` clean on both modified files.
- [ ] End-to-end verification against one of the three public NODD paths in #361 (e.g. `s3://unidata-nexrad-level2/2021/11/04/KLOT/KLOT20211104_174202_V06`) — recommended before merge.

### DIFF
--- a/tests/io/test_nexrad_level2.py
+++ b/tests/io/test_nexrad_level2.py
@@ -1955,6 +1955,114 @@ class TestIncompleteSweepParameter:
             call_kwargs = mock_open_sweeps.call_args[1]
             assert call_kwargs["incomplete_sweeps"] == {1}
 
+    # Regression cases for openradar/xradar#361. When upstream drops
+    # an interior cut, nex.data retains surviving sweeps under their
+    # original indices (sparse keys). The sweep-name list must follow
+    # those keys, not range(len(...)).
+    @pytest.mark.parametrize(
+        "present_keys,incomplete,mode,expected",
+        [
+            # Single interior gap (the #361 case, reduced)
+            (
+                [0, 1, 2, 4],
+                set(),
+                "drop",
+                ["sweep_0", "sweep_1", "sweep_2", "sweep_4"],
+            ),
+            (
+                [0, 1, 2, 4],
+                set(),
+                "pad",
+                ["sweep_0", "sweep_1", "sweep_2", "sweep_4"],
+            ),
+            # Gap + incomplete: drop removes both, pad removes only gap
+            (
+                [0, 1, 2, 4],
+                {2},
+                "drop",
+                ["sweep_0", "sweep_1", "sweep_4"],
+            ),
+            (
+                [0, 1, 2, 4],
+                {2},
+                "pad",
+                ["sweep_0", "sweep_1", "sweep_2", "sweep_4"],
+            ),
+            # Multiple interior gaps
+            (
+                [0, 1, 3, 4, 7],
+                set(),
+                "drop",
+                ["sweep_0", "sweep_1", "sweep_3", "sweep_4", "sweep_7"],
+            ),
+            # Trailing gap (MSG_5 declared 4 cuts, parser got 3)
+            (
+                [0, 1, 2],
+                set(),
+                "drop",
+                ["sweep_0", "sweep_1", "sweep_2"],
+            ),
+            # Sparse keys from unsorted dict iteration order
+            (
+                [4, 0, 2, 1],
+                set(),
+                "drop",
+                ["sweep_0", "sweep_1", "sweep_2", "sweep_4"],
+            ),
+        ],
+        ids=[
+            "single-gap-drop",
+            "single-gap-pad",
+            "gap-and-incomplete-drop",
+            "gap-and-incomplete-pad",
+            "multiple-gaps-drop",
+            "trailing-gap-drop",
+            "unsorted-keys-drop",
+        ],
+    )
+    def test_interior_sweep_gap_uses_actual_keys(
+        self, present_keys, incomplete, mode, expected
+    ):
+        mock_nex = MagicMock()
+        mock_nex.msg_5 = {"number_elevation_cuts": max(present_keys) + 2}
+        mock_nex.data = {k: None for k in present_keys}
+        mock_nex.incomplete_sweeps = incomplete
+        mock_nex.__enter__ = MagicMock(return_value=mock_nex)
+        mock_nex.__exit__ = MagicMock(return_value=False)
+
+        dummy_ds = xarray.Dataset(
+            {
+                "time": (
+                    "azimuth",
+                    np.array(
+                        ["2024-01-01T00:00:00", "2024-01-01T00:01:00"],
+                        dtype="datetime64[ns]",
+                    ),
+                ),
+                "latitude": 35.0,
+                "longitude": -97.0,
+                "altitude": 300.0,
+            },
+        )
+        mock_sweep_dict = OrderedDict([("sweep_0", dummy_ds), ("sweep_1", dummy_ds)])
+
+        with (
+            patch(
+                "xradar.io.backends.nexrad_level2.NEXRADLevel2File",
+                return_value=mock_nex,
+            ),
+            patch(
+                "xradar.io.backends.nexrad_level2.open_sweeps_as_dict",
+                return_value=mock_sweep_dict,
+            ) as mock_open_sweeps,
+        ):
+            open_nexradlevel2_datatree(
+                b"\x00" * 100,
+                reindex_angle=False,
+                incomplete_sweep=mode,
+            )
+            assert mock_open_sweeps.call_args[1]["sweeps"] == expected
+
 
 class TestPadModeReindexing:
     """Tests for the pad-mode reindex pipeline using real NEXRAD data.
--- a/xradar/io/backends/nexrad_level2.py
+++ b/xradar/io/backends/nexrad_level2.py
@@ -2091,8 +2091,14 @@ def open_nexradlevel2_datatree(
 
     # Single metadata read for sweep count, completeness, and elevation data
     with NEXRADLevel2File(filename_or_obj, loaddata=False) as nex:
-        act_sweeps = len(nex.msg_31_data_header)
+        # Reading incomplete_sweeps also triggers data_header parsing,
+        # populating nex.data. Must run before sorted(nex.data) below.
         incomplete = nex.incomplete_sweeps
+        # Use the sweep indices actually present, not range(len(...)):
+        # upstream-dropped interior sweeps leave sparse keys (e.g.
+        # [0..9, 11]) that would otherwise KeyError downstream. See #361.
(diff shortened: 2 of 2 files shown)


kmuehlbauer (approved):
This fix LGTM. Not looked deeper into the test, though.

PR #372, opened 2026-04-30
FIX: add id-token write to testpypi CI publisher

### LINKED ISSUE #364: CI: main-branch TestPyPI upload fails due to missing id-token: write permission
## Summary

The `test build distribution for testpypi` job in [`.github/workflows/ci.yml`](https://github.com/openradar/xradar/blob/main/.github/workflows/ci.yml) fails on every push to `main` at the **Upload to TestPyPI (Trusted Publisher)** step with:

> Trusted publishing exchange failure:
> OpenID Connect token retrieval failed: GitHub: missing or insufficient OIDC token permissions, the `ACTIONS_ID_TOKEN_REQUEST_TOKEN` environment variable was unset

Most recent failing run: https://github.com/openradar/xradar/actions/runs/24688919432/job/72206306374 (triggered by the #362 merge; the same failure has occurred on prior main pushes).

## Root cause

The workflow declares workflow-level `permissions: contents: read` at the top of `ci.yml` (correct least-privilege default), but the `test_build_distribution_testpypi` job never overrides it. `pypa/gh-action-pypi-publish` with **Trusted Publishing** (no API token, OIDC-based) needs `id-token: write` — with only `contents: read`, the OIDC token exchange can't happen.

The upload step is already gated with `if: ${{ github.event_name == 'push' }}`, so PR builds aren't affected; only main-branch pushes fail at this step.

## Proposed fix

Add a job-level `permissions:` block to that one job (keeps least-privilege — no other job gets `id-token: write`):

```yaml
  test_build_distribution_testpypi:
    name: test build distribution for testpypi
    needs: [lint, build_0]
    runs-on: ubuntu-latest
    permissions:
      contents: read
      id-token: write
    defaults:
      run:
        shell: bash -l {0}
```

This matches the ["Trusted Publishing" docs](https://docs.pypi.org/trusted-publishers/using-a-publisher/) and the [pypa/gh-action-pypi-publish README](https://github.com/pypa/gh-action-pypi-publish#trusted-publishing).

## Open questions for maintainers

- Should we also apply the same fix to `test_build_distribution_pypi` (PyPI upload, if/when it's enabled for real releases)?
- Is there a reason the testpypi upload is desirable on every main push vs. only on tagged releases? If main-push publishes are not intentional, the cleaner fix is to drop the `if: ${{ github.event_name == 'push' }}` gate in favor of `if: startsWith(github.ref, 'refs/tags/')`.

Happy to open a PR with whichever direction maintainers prefer — flagging it here first per the draft AI Usage Policy in #363 (CI / workflow changes should start with an issue).

CC: @kmuehlbauer @mgrover1

kmuehlbauer:
- [x] Closes #364

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,8 @@ jobs:
     name: test build distribution for testpypi
     needs: [lint, build_0]
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
We'll see this working only after the merge into main. I followed textbook, setting ``id-token: write`` should enable it proper.

kmuehlbauer:
Fingers crossed...

kmuehlbauer:
Worked, back in business :tada:

PR #374, opened 2026-05-03
FIX: open_store_coordinates handles interior sweep gaps (#366)

### LINKED ISSUE #366: open_store_coordinates still positionally indexes msg_31_header — #362 didn't fully close #361
## Summary

[#362](https://github.com/openradar/xradar/pull/362) fixed the **top-level** iteration in `open_nexradlevel2_datatree` to use the actual `nex.data` keys instead of `range(act_sweeps)`. That fix is correct but **partial**: a second call site still positionally indexes into the compacted `msg_31_header` list, so any NEXRAD Level-II volume with an interior sweep-index gap (one cut dropped upstream, remaining cuts keeping their original labels) still fails at DataTree construction — now with `IndexError` instead of the original `KeyError`.

Root cause is `nexrad_level2.py:1755` inside `open_store_coordinates` (the per-sweep entrypoint called downstream of the fixed sweep-list logic):

```python
def open_store_coordinates(self):
    msg_31_header = self.root.msg_31_header[self._group]
    ...
```

When the file has 11 present sweeps with keys `[0, 1, …, 9, 11]`, `self._group` for the last sweep is `11`, but `msg_31_header` has been compacted to length 11, so `msg_31_header[11]` → `IndexError: list index out of range`. `msg_31_data_header` in the same state keeps the full declared length (12, with a placeholder at the gap position), so the adjacent line `self.root.msg_31_data_header[self._group]` happens to succeed — which is part of why this wasn't caught under MagicMock-based unit tests.

## Reproduction (public NODD file, ~5 seconds)

```python
import fsspec, xradar as xd
from xradar.io.backends.nexrad_level2 import NEXRADLevel2File

url = "s3://unidata-nexrad-level2/2021/11/04/KLOT/KLOT20211104_174202_V06"
with fsspec.open(url, mode="rb", anon=True) as fh:
    buf = fh.read()

nex = NEXRADLevel2File(buf)
print("MSG_5 cuts        :", nex.get_msg_5_data()["number_elevation_cuts"])
print("len(msg_31_header) :", len(nex.msg_31_header))
print("sorted(data)      :", sorted(nex.data))

# MSG_5 cuts        : 12
# len(msg_31_header) : 11        <- compacted (no placeholder at the gap position)
# sorted(data)      : [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11]

dt = xd.io.open_nexradlevel2_datatree(buf)
# -> IndexError: list index out of range   at open_store_coordinates, line 1755
```

xradar version: `0.12.0` (just released with #362 merged).

Four documented entry paths all fail the same way on this file:

| Approach | Result |
|---|---|
| `open_nexradlevel2_datatree(buf)` (default) | `IndexError: list index out of range` |
| `sweep=[0,1,…,9,11]` (preserve labels, explicit list) | `IndexError` — still hits line 1755 |
| `sweep=list(range(11))` (renumber to contiguous) | `KeyError: 10` — `nex.data[10]` doesn't exist |
| `incomplete_sweep="pad"` | `IndexError` |

## Internal state inconsistency

After `nex.incomplete_sweeps` (and implicitly `nex.data`) have been parsed on the reproduction file:

```
sorted(nex.data)              = [0, 1, 2, …, 9, 11]         (len 11, actual present keys)
nex.msg_31_header             = list, length 11              (compacted — NO placeholder at idx 10)
nex.msg_31_data_header        = list, length 12              (keeps declared length with placeholder)
```

So `self.root.msg_31_data_header[self._group]` works for any label in `[0..11]`, but `self.root.msg_31_header[self._group]` only works when labels happen to equal compact indices — i.e. when there is no gap. Any code path that positionally indexes into `msg_31_header` using `self._group` will break on gap files.

## Proposed fix (validated locally)

```diff
 def open_store_coordinates(self):
-    msg_31_header = self.root.msg_31_header[self._group]
+    # self._group is the sweep LABEL (e.g. 11 when one interior cut was
+    # dropped upstream and surviving sweeps kept their original indices).
+    # msg_31_header is compacted (length = len(nex.data)), so translate
+    # label -> compact position. msg_31_data_header keeps the declared
+    # length, so it stays indexed by label.
+    present_keys = sorted(self.root.data)
+    pos = present_keys.index(self._group)
+    msg_31_header = self.root.msg_31_header[pos]
     msg_31_data_header = self.root.msg_31_data_header[self._group]
```

`msg_5["elevation_data"][self._group]` a few lines below also uses the declared length (same as `msg_31_data_header`), so it stays indexed by label — no change needed there. Healthy non-gap files are unaffected because `present_keys.index(self._group) == self._group` whenever the keys are contiguous.

## Validation — three real NODD files

Applied the patch locally to the installed `xradar==0.12.0` and opened each file end-to-end:

| File | MSG_5 cuts | `sorted(data)` | Result with patch |
|---|---|---|---|
| `s3://unidata-nexrad-level2/2021/11/04/KLOT/KLOT20211104_174202_V06` | 12 | `[0..9, 11]` | **11 sweeps** — `sweep_0..sweep_9`, `sweep_11`, no `sweep_10`; angles match |
| `s3://unidata-nexrad-level2/2021/11/19/KLOT/KLOT20211119_162142_V06` |  8 | `[0, 2..7]`  | **7 sweeps** — `sweep_0`, `sweep_2..sweep_7`, no `sweep_1`; angles match |
| `s3://unidata-nexrad-level2/2022/04/24/KLOT/KLOT20220424_151601_V06` | 18 | `[0..10, 12..17]` | **17 sweeps** — `sweep_0..sweep_10`, `sweep_12..sweep_17`, no `sweep_11`; angles match |

Per-sweep elevation-angle comparison for the first file (every present sweep matches the MSG_5 schedule within 0.2°):

```
cut    declared       actual        match
 0     0.4833984375   0.50537109375   OK
 1     0.4833984375   0.3955078125    OK
 2     0.87890625     0.7965087890625 OK
 3     0.87890625     0.8349609375    OK
 4     1.318359375    1.31011962890625 OK
 5     1.318359375    1.241455078125  OK
 6     1.8017578125   1.77978515625   OK
 7     2.4169921875   2.2247314453125 OK
 8     3.1201171875   3.00201416015625 OK
 9     3.9990234375   3.8397216796875 OK
10     5.09765625     MISSING         -    <- dropped by upstream, no placeholder in data
11     6.416015625    6.339111328125  OK   <- kept its original index 11
```

So the data carried by the file is valid — every present cut reads back with the correct elevation — the only thing that prevents it from assembling into a DataTree is the positional-index in `open_store_coordinates`.

Healthy non-gap files tested (e.g. `s3://unidata-nexrad-level2/2025/03/14/KLOT/KLOT20250314_233157_V06`, 12 sweeps, keys `[0..11]`) produce identical output before and after the patch.

## Suggested regression test

The PR #362 tests mock `open_sweeps_as_dict` (`with patch("...open_sweeps_as_dict")`), so the test flow never reaches `open_store_coordinates` and the bug couldn't surface. A fixture that mocks one level deeper — or, simpler, exercises the real NODD paths above end-to-end — catches this:

```python
import fsspec, xradar as xd, pytest

@pytest.mark.parametrize("url, expected_keys", [
    ("s3://unidata-nexrad-level2/2021/11/04/KLOT/KLOT20211104_174202_V06",
     [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11]),
    ("s3://unidata-nexrad-level2/2021/11/19/KLOT/KLOT20211119_162142_V06",
     [0, 2, 3, 4, 5, 6, 7]),
    ("s3://unidata-nexrad-level2/2022/04/24/KLOT/KLOT20220424_151601_V06",
     [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 13, 14, 15, 16, 17]),
])
def test_interior_gap_file_decodes_end_to_end(url, expected_keys):
    with fsspec.open(url, mode="rb", anon=True) as fh:
        buf = fh.read()
    dt = xd.io.open_nexradlevel2_datatree(buf)
    sweep_names = [c for c in dt.children if c.startswith("sweep_")]
    assert sweep_names == [f"sweep_{k}" for k in expected_keys]
```

If hitting real S3 in CI is undesirable, the same shape can be reproduced by constructing a synthetic `NEXRADLevel2File` state where `msg_31_header` has length N, `msg_31_data_header` has length N+1, and `data` has N keys with an interior gap — but that's more fragile than the three-file integration test against stable NODD paths.

## Scope note

This is narrowly about the interior-sweep-gap bug pattern — files that xradar 0.11.x would reject with `KeyError` and 0.12.0 still rejects with `IndexError`. It does NOT address other "CORRUPTED"-looking files with different pathologies (e.g. truncated files with unreadable MSG_5, or files where different sweeps within the same volume have mismatched ray counts — those produce `ValueError: conflicting sizes for dimension 'azimuth'` and are genuine source-level corruption, not a decoder issue).

Happy to open a follow-up PR with the diff above + the regression test.

aladinor:
## Summary

Follow-up to #362. That PR fixed the top-level iteration in `open_nexradlevel2_datatree` but `NexradLevel2Store.open_store_coordinates` still positionally indexed `self.root.msg_31_header` by sweep label, so volumes with an interior sweep gap fell over with `IndexError` (instead of the original `KeyError`).

`msg_31_header` is compacted to `len(nex.data)`; `msg_31_data_header` keeps the declared length with a placeholder. The fix translates label → compact position via `sorted(self.root.data).index(self._group)` and applies it at both call sites in `open_store_coordinates`.

Closes #366.

## Repro (from the issue)

```python
import fsspec, xradar as xd
url = "s3://unidata-nexrad-level2/2021/11/04/KLOT/KLOT20211104_174202_V06"
with fsspec.open(url, mode="rb", anon=True) as fh:
    buf = fh.read()
xd.io.open_nexradlevel2_datatree(buf)
# Before: IndexError: list index out of range  at open_store_coordinates
# After:  DataTree with sweep_0..sweep_9, sweep_11
```

## Test plan

- [x] New `TestInteriorGapCoordinates` exercises `open_store_coordinates` directly (22 cases: 11 present labels × 2 `msg_5.elevation_data` states)
- [x] Per-position elevation values in the synthetic header verify the label→position translation, not just absence of `IndexError`
- [x] Confirmed all 22 cases fail on the unfixed code with the exact `IndexError` from the issue, and pass with the fix
- [x] Full `tests/io/test_nexrad_level2.py` runs clean (126 passed)
- [x] `black --check` and `ruff check` clean

## Notes

- The #362 tests mock `open_sweeps_as_dict`, which short-circuits before reaching the buggy path — that's how the bug shipped. The new tests drop one layer deeper to a synthetic `NexradLevel2Store`, so the actual call site is exercised.
- Real-NODD integration coverage (the three KLOT URLs from the issue) is intentionally left out of this PR; it can land later via an `open-radar-data` fixture (tracked separately).

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -2,6 +2,7 @@
 
 ## Development
 
+* FIX: ``open_nexradlevel2_datatree`` decodes volumes with interior sweep-index gaps end-to-end — translate sweep label → compact position in ``NexradLevel2Store.open_store_coordinates`` so the per-sweep entrypoint stops positionally indexing the compacted ``msg_31_header`` (follow-up to {pull}`362`) ({issue}`366`, {pull}`374`) by [@aladinor](https://github.com/aladinor)
 * DOC: Add projection comparison and cartopy map examples to ``Georeference_TargetCRS`` notebook by [@syedhamidali](https://github.com/syedhamidali)
 * DOC: Add full-form descriptions for all supported radar formats in README.md by [@syedhamidali](https://github.com/syedhamidali)
 * ADD: India Meteorological Department (IMD) radar NetCDF reader (``IMDBackendEntrypoint``, ``open_imd_datatree``). IMD stores one sweep per file; ``open_imd_datatree`` accepts a single file or a list of files to assemble a multi-sweep volume ({issue}`368`, {pull}`367`) by [@syedhamidali](https://github.com/syedhamidali)
--- a/docs/notebooks/nexrad_read_chunks.md
+++ b/docs/notebooks/nexrad_read_chunks.md
@@ -73,11 +73,9 @@ chunk_paths = []
 try:
     fs = fsspec.filesystem("s3", anon=True)
     volumes = sorted(fs.ls("unidata-nexrad-level2-chunks/KABR/"))
-    # Real-time chunks arrive incrementally and age out one at a time, so the
-    # most recent volume may be mid-flight (no S yet) and recent ones may have
-    # lost their S. Walk newest -> oldest and require BOTH the S (start) and
-    # E (end) chunks so we know the volume is complete and decodable.
-    for vol in reversed(volumes):
+    # Newest volume may be mid-flight (no E); aging volumes may have lost
+    # their S. Walk back until we find one with both chunks present.
+    for vol in list(reversed(volumes))[:10]:
         candidate_paths = sorted(fs.ls(vol))
         suffixes = {p.rsplit("-", 1)[-1].rsplit("_", 1)[-1] for p in candidate_paths}
         if "S" in suffixes and "E" in suffixes:
--- a/docs/notebooks/nexrad_read_chunks.md
+++ b/docs/notebooks/nexrad_read_chunks.md
@@ -73,13 +73,16 @@ chunk_paths = []
 try:
     fs = fsspec.filesystem("s3", anon=True)
     volumes = sorted(fs.ls("unidata-nexrad-level2-chunks/KABR/"))
-    if volumes:
-        candidate_paths = sorted(fs.ls(volumes[-1]))
-        # The S (start) chunk carries the volume header (AR2V prefix); without
-        # it the concatenated bytes can't be decoded. Real-time chunks age out,
-        # so the most recent volume may have only I/E chunks left.
-        if any(p.endswith("_S") for p in candidate_paths):
+    # Real-time chunks arrive incrementally and age out one at a time, so the
+    # most recent volume may be mid-flight (no S yet) and recent ones may have
+    # lost their S. Walk newest -> oldest and require BOTH the S (start) and
+    # E (end) chunks so we know the volume is complete and decodable.
+    for vol in reversed(volumes):
+        candidate_paths = sorted(fs.ls(vol))
+        suffixes = {p.rsplit("-", 1)[-1].rsplit("_", 1)[-1] for p in candidate_paths}
+        if "S" in suffixes and "E" in suffixes:
             chunk_paths = candidate_paths
+            break
 except Exception:
     pass
 
--- a/docs/notebooks/nexrad_read_chunks.md
+++ b/docs/notebooks/nexrad_read_chunks.md
@@ -74,7 +74,12 @@ try:
     fs = fsspec.filesystem("s3", anon=True)
     volumes = sorted(fs.ls("unidata-nexrad-level2-chunks/KABR/"))
     if volumes:
-        chunk_paths = sorted(fs.ls(volumes[-1]))
+        candidate_paths = sorted(fs.ls(volumes[-1]))
+        # The S (start) chunk carries the volume header (AR2V prefix); without
+        # it the concatenated bytes can't be decoded. Real-time chunks age out,
+        # so the most recent volume may have only I/E chunks left.
+        if any(p.endswith("_S") for p in candidate_paths):
+            chunk_paths = candidate_paths
 except Exception:
     pass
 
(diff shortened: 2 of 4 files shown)


kmuehlbauer (approved):
Thanks @aladinor!

PR #383, opened 2026-05-23
ENH: extract nominal ray angle by default for ODIM backend

egouden:
EXPERIMENTAL PR

This PR was produced together with DeepSeek V4 Flash Free (OpenCode Zen). I provided guidance on the goal and expectations. The main implementation has been made by me 6 months ago without AI. I needed AI help for review and the new kwargs and tests. I did not check these as my experience is limited.

Our summary:

Adds azimuth_mode option (nominal/per_ray) to ODIM backend with fallback for non-standard nrays.

Our rationale:

Nominal azimuth (uniform sweep from startazA[0]) is the default because it's more robust: per-ray startazA/stopazA often have antenna jitter that isn't meaningful, and most downstream algorithms (gridding, interpolation) assume uniform spacing anyway. The old per-ray method is preserved as a fallback for non-standard nrays, and users can opt into it explicitly via azimuth_mode="per_ray" if they need the raw midpoints.

EXPERIMENTAL REVIEW

In depth review by mistral-medium, a bigger european model

In-Depth Review: Commit fe06f12
Summary: Adds azimuth_mode parameter to ODIM backend, defaulting to "nominal" mode which extracts uniform ray angles for standard sweep configurations (180/360/720 rays) instead of calculating per-ray midpoints.
The Problem Being Solved
ODIM files store azimuth information in two ways:
 1. 
Per-ray: startazA and stopazA arrays defining start/stop angles for each ray
 2. 
Nominal/uniform: Standard PPI sweeps with uniformly spaced rays (180, 360, 720 rays)
The previous implementation always calculated per-ray midpoints from startazA/stopazA, which could introduce small inconsistencies or unnecessary complexity for standard uniform sweeps. Many radar files, especially from operational networks, use standard sweep patterns where azimuth angles are perfectly uniform.

Implementation Assessment
✅ Strengths:
 1. 
Backward Compatible: Default behavior now prefers nominal angles, but:
• 
Falls back to per-ray calculation for non-standard ray counts
• 
Explicit azimuth_mode="per_ray" restores old behavior
• 
All existing code continues to work
 2. 
Well-Architected: The azimuth_mode parameter is properly threaded through:
• 
_OdimH5NetCDFMetadata.__init__
• 
OdimSubStore.__init__
• 
OdimStore.__init__ and .open()
• 
OdimBackendEntrypoint.open_dataset()
 3. 
Sensible Defaults:
• 
Standard ray counts (180, 360, 720) cover most operational radars
• 
Starting angle calculated from first startazA value or astart attribute
• 
Graceful fallback with clear warnings
 4. 
Good Error Handling:
• 
Warnings for missing startazA/astart (defaults to 0)
• 
Warnings for non-standard ray counts
• 
try/except blocks for missing attributes
⚠️ Considerations:
 1. 
Magic Numbers: The hardcoded [180, 360, 720] list is pragmatic but inflexible. Some radars use 450, 900, or other counts. Consider making this configurable.
 2. 
Angle Rounding Logic:
astart = np.round(astart / (ascale / 2)) * ascale / 2
This snaps the start angle to the nearest half-step. Good for handling minor discrepancies, but could be documented better.
 3. 
Warning Message: "Unexpected number of rays" might be confusing. Consider "Non-standard number of rays" or similar.
 4. 
Dtype: The nominal azimuth uses default numpy dtype (float64), while _get_azimuth_where used float32. Minor inconsistency.

Overall Verdict: Good Implementation (8.5/10)
The implementation is well-designed, backward-compatible, and solves a real problem. The choice of defaulting to nominal mode is reasonable for the target use case (ODIM operational radar data).
Minor improvements for future consideration:
• 
Make standard ray counts configurable
• 
Clarify warning messages
• 
Document the angle snapping logic
• 
Consider dtype consistency
The commit represents a thoughtful enhancement that improves data consistency for the most common use cases while maintaining full backward compatibility.

### DIFF
--- a/xradar/io/backends/odim.py
+++ b/xradar/io/backends/odim.py
@@ -100,8 +100,6 @@ def _get_azimuth_how(how):
     startaz = how["startazA"]
     stopaz = how.get("stopazA", False)
     if stopaz is False:
-        # stopazA missing
-        # create from startazA
         stopaz = np.roll(startaz, -1)
         stopaz[-1] += 360
     zero_index = np.where(stopaz < startaz)
@@ -114,6 +112,27 @@ def _get_azimuth_how(how):
 def _get_azimuth_where(where):
     res = 360.0 / where["nrays"]
     return np.arange(res / 2.0, 360.0, res, dtype="float32")
+
+
+def _get_azimuth_nominal(nrays, how):
+
+    if nrays not in [180, 360, 450, 720, 900]:
+        warnings.warn(f"xradar: Unexpected number of rays ({nrays})")
+        raise ValueError
+
+    ascale = 360 / nrays
+    try:
+        astart = how["startazA"][0]
+        astart = min(astart, astart - 360, key=abs)
+        astart = np.round(astart / (ascale / 2)) * ascale / 2
+    except (KeyError, TypeError, AttributeError):
+        try:
+            astart = how["astart"]
+        except KeyError:
+            warnings.warn("xradar: No startazA or astart found, using 0 as default.")
+            astart = 0
+    azimuth = np.arange(astart + ascale / 2, 360, ascale)
+    return azimuth
 
 
 def _get_fixed_dim_and_angle(where):
@@ -407,11 +426,18 @@ class _OdimH5NetCDFMetadata(_H5NetCDFMetadata):
         h5netcdf filehandle.
     group : str
         odim group to acquire
+    azimuth_mode : str
+        ``"nominal"`` (default) for uniform sweep from first ``startazA``,
+        ``"per_ray"`` for per-ray midpoint from ``startazA``/``stopazA``.
 
     Returns
     -------
     object : metadata object
     """
+
+    def __init__(self, fileobj, group, azimuth_mode="nominal"):
+        super().__init__(fileobj, group)
+        self._azimuth_mode = azimuth_mode
 
     @property
     def ds_what(self):
@@ -461,10 +487,18 @@ class _OdimH5NetCDFMetadata(_H5NetCDFMetadata):
 
     @property
     def _azimuth(self):
-        try:
-            azimuth = _get_azimuth_how(self.how)
-        except (AttributeError, KeyError, TypeError):
-            azimuth = _get_azimuth_where(self.where)
+        nrays = self.where["nrays"]
+        if self._azimuth_mode == "nominal":
+            try:
+                azimuth = _get_azimuth_nominal(nrays, self.how)
+            except ValueError:
+                warnings.warn("xradar: Using per ray real azimuth")
+                self._azimuth_mode = "per_ray"
+        if self._azimuth_mode == "per_ray":
+            try:
+                azimuth = _get_azimuth_how(self.how)
+            except (AttributeError, KeyError, TypeError):
+                azimuth = _get_azimuth_where(self.where)
         return Variable((self.dim0,), azimuth, get_azimuth_attrs())
 
     @property
@@ -607,6 +641,7 @@ class OdimSubStore(AbstractDataStore):
         store,
         group=None,
         lock=False,
+        azimuth_mode="nominal",
     ):
         if not isinstance(store, OdimStore):
             raise TypeError(
@@ -619,11 +654,14 @@ class OdimSubStore(AbstractDataStore):
         self._filename = store.filename
         self.is_remote = is_remote_uri(self._filename)
         self.lock = ensure_lock(lock)
+        self._azimuth_mode = azimuth_mode
 
     @property
     def root(self):
         with self._manager.acquire_context(False) as root:
-            return _OdimH5NetCDFMetadata(root, self._group.lstrip("/"))
+            return _OdimH5NetCDFMetadata(
+                root, self._group.lstrip("/"), azimuth_mode=self._azimuth_mode
+            )
 
     def _acquire(self, needs_lock=True):
         with self._manager.acquire_context(needs_lock) as root:
@@ -663,7 +701,7 @@ class OdimSubStore(AbstractDataStore):
 class OdimStore(AbstractDataStore):
     """Store for reading ODIM dataset groups via h5netcdf."""
 
-    def __init__(self, manager, group=None, lock=False):
+    def __init__(self, manager, group=None, lock=False, azimuth_mode="nominal"):
         if isinstance(manager, (h5netcdf.File, h5netcdf.Group)):
             if group is None:
                 root, group = find_root_and_group(manager)
@@ -683,6 +721,7 @@ class OdimStore(AbstractDataStore):
         self.lock = ensure_lock(lock)
         self._substore = None
         self._need_time_recalc = False
+        self._azimuth_mode = azimuth_mode
 
     @classmethod
     def open(
@@ -695,6 +734,7 @@ class OdimStore(AbstractDataStore):
         invalid_netcdf=None,
         phony_dims=None,
         decode_vlen_strings=True,
+        azimuth_mode="nominal",
     ):
         if isinstance(filename, bytes):
             raise ValueError(
@@ -717,7 +757,7 @@ class OdimStore(AbstractDataStore):
                 lock = False
 
         manager = CachingFileManager(h5netcdf.File, filename, mode=mode, kwargs=kwargs)
-        return cls(manager, group=group, lock=lock)
+        return cls(manager, group=group, lock=lock, azimuth_mode=azimuth_mode)
 
     @property
     def filename(self):
@@ -741,6 +781,7 @@ class OdimStore(AbstractDataStore):
                             self,
                             group=group,
                             lock=self.lock,
+                            azimuth_mode=self._azimuth_mode,
                         )
                         for group in subgroups
                     ]
@@ -786,6 +827,9 @@ class OdimBackendEntrypoint(BackendEntrypoint):
         If True, fixes erroneous second angle data. Defaults to ``False``.
     site_as_coords : bool
         Attach radar site-coordinates to Dataset, defaults to ``True``.
+    azimuth_mode : str
+        ``"nominal"`` (default) for uniform sweep from first ``startazA``,
+        ``"per_ray"`` for per-ray midpoint from ``startazA``/``stopazA``.
     kwargs : dict
         Additional kwargs are fed to :py:func:`xarray.open_dataset`.
     """
@@ -813,6 +857,7 @@ class OdimBackendEntrypoint(BackendEntrypoint):
         reindex_angle=False,
         fix_second_angle=False,
         site_as_coords=True,
+        azimuth_mode="nominal",
     ):
         if isinstance(filename_or_obj, io.IOBase):
             filename_or_obj.seek(0)
@@ -824,6 +869,7 @@ class OdimBackendEntrypoint(BackendEntrypoint):
             invalid_netcdf=invalid_netcdf,
             phony_dims=phony_dims,
             decode_vlen_strings=decode_vlen_strings,
+            azimuth_mode=azimuth_mode,
         )
 
         store_entrypoint = StoreBackendEntrypoint()
--- a/tests/io/test_odim.py
+++ b/tests/io/test_odim.py
@@ -57,6 +57,126 @@ def test_get_azimuth_where(nrays):
     assert udiff[0] == 360.0 / nrays
 
 
+def test_odim_azimuth_nominal(odim_file):
+    import xarray as xr
+
+    ds = xr.open_dataset(odim_file, engine="odim", group="sweep_0")
+    azimuth = ds.azimuth.values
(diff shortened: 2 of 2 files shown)


rcjackson:
I would say that your PR message needs to be human-generated. LLMs tend to be extremely verbose in their messages, to the point of confusion. Then, I think it shows that you understand your PR and helps us as maintainers understand it better.

egouden:
Thank you for the comment.

I updated the PR for clarity. I stressed this is team effort with AI. I take responsibility for the text.

The relatively long review, which seems valid, is actually not part of the PR. It is for discussion purposes related to AI usage in review.

xradar/io/backends/odim.py
```diff
@@ -96,26 +96,6 @@ def _calculate_angle_res(dim):
     return np.round(np.nanmean(angle_diff_wanted), decimals=2)
 
 
-def _get_azimuth_how(how):
```
kmuehlbauer, line 99: Please leave these two functions here, instead of copying around.

xradar/io/backends/odim.py
```diff
@@ -461,10 +466,32 @@ def a1gate(self):
 
     @property
     def _azimuth(self):
-        try:
-            azimuth = _get_azimuth_how(self.how)
-        except (AttributeError, KeyError, TypeError):
-            azimuth = _get_azimuth_where(self.where)
+        nrays = self.where["nrays"]
+        if self._azimuth_mode == "nominal" and nrays in [180, 360, 720]:
```
kmuehlbauer: Can we move this into a dedicated function, like `_get_azimuth_nominal()`. 

I'd also move the check for `[180, 360, 720]` inside that function. Why do we want to restrict this to these numbers? ODIM_H5 doesn't mention anything related.

It looks like this is implementation detail and only works if all rays of a certain layout are available. Otherwise it will fall back.

xradar/io/backends/odim.py
```diff
@@ -461,10 +466,32 @@ def a1gate(self):
...
+            ascale = 360 / nrays
+            try:
+                astart = self.how["startazA"][0]
+                astart = min(astart, astart - 360, key=abs)
+                astart = np.round(astart / (ascale / 2)) * ascale / 2
```
kmuehlbauer: Good, this gets us a value what `astart` is described in ODIM_H5:

_Azimuthal offset from 0◦ of the start of the first ray in the sweep. This value is positive where the gate starts clockwise after 0◦, and it will be negative if it starts before 0◦. In either case, the value must be no larger than half a ray’s width._

xradar/io/backends/odim.py
```diff
@@ -461,10 +466,32 @@ def a1gate(self):
...
+            except (KeyError, TypeError, AttributeError):
+                try:
+                    astart = self.how["astart"]
```
kmuehlbauer: So, why falling back to this here? According to the standard `astart` is optional whereas `startazA` is mandatory. 

I've skimmed the available standards (https://www.eumetnet.eu/observations/weather-radar-network/, and click on "Read More" on OPERA Data Information Model) and `startazA` is recommended since 2.1 and mandatory since 2.4. `astart` was first mentioned in 2.2.

Maybe check for `astart` first (easy) and fallback to `startAzA`?

xradar/io/backends/odim.py
```diff
@@ -461,10 +466,32 @@ def a1gate(self):
...
+                except KeyError:
+                    warnings.warn(
+                        "xradar: No startazA or astart found, using 0 as default."
+                    )
+                    astart = 0
+            azimuth = np.arange(astart + ascale / 2, 360, ascale)
```
kmuehlbauer: Do we need to check if the correct number of azimuth angles is created here? Or could it happen, that `len(azimuth) != nrays` for some reason?

xradar/io/backends/odim.py
```diff
@@ -461,10 +466,32 @@ def a1gate(self):
...
+        else:
+            if nrays not in [180, 360, 720]:
```
kmuehlbauer: Move that nrays check into above suggested function.

xradar/io/backends/odim.py
```diff
@@ -461,10 +466,32 @@ def a1gate(self):
...
+                warnings.warn(
+                    f"xradar: Unexpected number of rays ({nrays}), "
+                    "falling back to per-ray azimuth calculation."
+                )
+            try:
+                azimuth = _get_azimuth_how(self.how)
+            except (AttributeError, KeyError, TypeError):
+                azimuth = _get_azimuth_where(self.where)
```
kmuehlbauer, line 501: In light of the new nominal azimuth approach it might be time to refactor this try/except. Can be done in a later PR.

xradar/io/backends/odim.py
```diff
@@ -786,6 +820,9 @@ class OdimBackendEntrypoint(BackendEntrypoint):
         If True, fixes erroneous second angle data. Defaults to ``False``.
     site_as_coords : bool
         Attach radar site-coordinates to Dataset, defaults to ``True``.
+    azimuth_mode : str
```
kmuehlbauer, line 830: We need to preserve backwards compatibility. So `"nominal"` can't be the default.

xradar/io/backends/odim.py
```diff
@@ -824,6 +862,7 @@ def open_dataset(
             invalid_netcdf=invalid_netcdf,
             phony_dims=phony_dims,
             decode_vlen_strings=decode_vlen_strings,
+            azimuth_mode=azimuth_mode,
```
kmuehlbauer, line 872: So, after digging through this, we might do this a bit different.

What this PR aims to:

If `nrays in [180, 360, 720]` and `azimuth_mode="nominal"` return nominal azimuth angles.

This actually looks like a fast path of reindexing, which should be implemented anyway. How we could achieve this:

1. Use `reindex_angle=True` and return nominal angles IF the suggested fast path (extracting nominal rays from `startazA`/`astart`) would work. If the fast path does not work (missing/duplicate rays etc. pp) we go the usual way returning measured angles + reindexing.

2. If we want to switch behaviour and return nominal angles by default, we need to switch the default to `"reindex_angle=True"`

This would be a minimal change and would also speed up reading well-formed data when reindexing is requested.

kmuehlbauer (commented):
Thanks @egouden for this PR.

I'll just add my reasonings here a second time:

> _This actually looks like a fast path of reindexing, which should be implemented anyway. How we could achieve this:
> 
> 1. Use reindex_angle=True and return nominal angles IF the suggested fast path (extracting nominal rays from startazA/astart) would work. If the fast path does not work (missing/duplicate rays etc. pp) we go the usual way returning measured angles + reindexing.
> 
> 2. If we want to switch behaviour and return nominal angles by default, we need to switch the default to "reindex_angle=True"_

xradar/io/backends/odim.py
egouden: Done

xradar/io/backends/odim.py
egouden: A dedicated function has been created. The check for standard nrays is to catch malformed scans, which by definition are not nominal. There might be exotic well formed scans outside these numbers. I see no easy other way to identify malformed scan.

xradar/io/backends/odim.py
egouden: Yes and the issue is that astart is complex and can be wrong, hence the ODIM solution of mandatory startazA.

xradar/io/backends/odim.py
egouden: startazA is less ambiguous

xradar/io/backends/odim.py
egouden: No, it should always be correct, considering startzaA is within (lazy) specifications, of course, i.e. not below -0.75 or above 0.5.

xradar/io/backends/odim.py
egouden: Done

xradar/io/backends/odim.py
egouden: The API is the same. I do not think this will break code. I will repeat my rationale. Most users are expecting nominal angles. The radar data should be nominal by design. This is in my opinion enforced in the FM301 specification. This issue can be considered a flaw in ODIM format design (i.e. not having nominal astart mandatory). The old behavior is preserved as an option. This is one of the thing i would like to see fixed before 1.0. I would like to hear the opinion of other team members on this. Regarding your main comment, this is in my view not really about reindexing angles which is more a post processing approach (i used it too later in my workflow).

xradar/io/backends/odim.py
egouden: See my rationale above.

egouden:
The changes and comments are mine only. There was already a test with nominal 361 rays. I added one test with nominal rays but a double ray (with the help of grok 4.3).

xradar/io/backends/odim.py
kmuehlbauer: It's not about what most users expect, it's about what current users get when running their workflows. Currently users get measured angles. If we want to return nominal angles by default instead of measured angles we need to properly announce that change or better go through a deprecation cycle. One could claim that we are still in version `0.12` as of now, so we could change anything anytime. But I would really avoid that.

One more thing, would we want to have nominal angles for RHI, too?

xradar/io/backends/odim.py
kmuehlbauer: I was suggesting to return nominal angles per default (point 2 above) without adding the `azimuth_mode` `kwarg` at the backend level. 

Example 1 (perfect file, 360 rays):

This currently returns measured angles (default `reindex_angle=False`). To return `nominal` angles users have to use `reindex_angle=reindex_dict`. I'm proposing to use `reindex_angle=True` as a means of "try to return nominal angle" and return nominal angles directly from the odim reader.

Example 2 (broken file, 361 rays):

This currently returns measured angles. To return a repaired version with nominal angles users have to use `reindex_angle=reindex_dict`. I'm proposing to use `reindex_angle=True` as above and return an error in this case (malformed data).

Then, in a deprecation cycle we would switch from the default `reindex_angle=False` to `reindex_angle`=True. Then users always get nominal angles directly from the odim reader (fast path) side stepping the reindexing.

I'm open to change `reindex_angle` kwarg to a more fitting name.

xradar/io/backends/odim.py
egouden: I understand the concern for stability. I think fixing this properly  would have a positive impact on both old and new users.

To be consistent with FM301, all angles should be nominal. Are there other readers returning measured angles?

kmuehlbauer: Most readers return measured angles if they are available. I'm on board with returning nominal angles by default. But we should keep behaviour consistent and introduce the change via deprecation cycle. This would also have the benefit of properly announcing this new scheme.

We could start announcing it like this:

`reindex_angle=False`: DeprecationWarning("Default angle output will change in a future version, nominal angles (`reindex_angle="nominal"`) will become the default. To silence this warning and keep current measured angles, set `reindex_angle="raw"` now.")"

Then we implement the fast path along your implementation for `reindex_angle="nominal"`. The error message for non-conforming sweeps should state that the user has to either provide a fitting reindex_dict or use `reindex_angle="raw"` to return raw measured angles.

I think this is the easiest and fastest way without cluttering the API with yet another keyword. 

Let's assume "nominal_angles" as our new default. Any reindex operation which repairs/fixes issues with the data will also return nominal angles. So, we could just alias the current `reindex_angle` with eg. just `angle`.

```
angle="nominal"  # get nominal angle
angle="raw"  # get raw (measured) angle
angle=dict(direction=1, angle_res=1., start_angle=0, stop_angle=360)  # get reindexed nominal angles
```
WDYT?

egouden: This sounds like a great plan! I am always in favor to keep it simple. I guess it is more efficient if you take over from this point in a new integrated PR? I can help of course.

kmuehlbauer: I'll try to re-use as much of your implementation as possible and have a PR ready soon.

xradar/io/backends/odim.py
egouden: I understand now. nominal_angle=true?

kmuehlbauer: I'm still considering the naming, `reindex_angle` does not fit any more, `azimuth_mode` does not cover for RHI, `angle_mode` does not reflect the possible mapping input. Pure `angle` might be a bit too implicit. I'm tending to use `angle_spec`.

My current idea is to move along those lines:

```
angle_spec = "nominal"  # try to return nominal angles, raise if not possible
angle_spec = "raw"  # just return mid point angles from whatever data is in the file
angle_spec = "auto"  # `extract_angle_parameters`, reindex and return nominal angles
angle_spec = reindex_mapping  # reindex with given mapping, return nominal angles
```

Still checking possible best data flow...

PR #385, opened 2026-06-04
ENH: When importing ODIM files, extract the global attributes ‘wmo__i…

chfer:
## Summary

This PR extends the ODIM backend to extract additional global attributes from the `/what/source` metadata field when importing ODIM files.

The extracted values are mapped onto the following FM301 global attributes in the xradar memory model:
- `wmo__id`
- `site_name`
- `node`
- `wmo__originating_centre`
- `wmo__wsi`

## Details

The ODIM `/what/source` string is parsed as key-value metadata and used to populate dataset global attributes during import.

This change also handles source strings containing non-ASCII station names encoded through surrogate-escaped UTF-8 byte sequences, so values such as for example `Sürgavere` are preserved correctly.

## Tests

Added coverage to verify:
- parsing of `/what/source` into global attributes
- correct recovery of UTF-8 text from surrogate-escaped source strings

### DIFF
--- a/tests/io/test_odim.py
+++ b/tests/io/test_odim.py
@@ -9,6 +9,7 @@ ported from wradlib
 
 from contextlib import nullcontext
 
+import h5netcdf
 import numpy as np
 import pytest
 from xarray import DataTree, open_dataset, open_mfdataset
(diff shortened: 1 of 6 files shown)


kmuehlbauer:
Thanks @chfer, that looks like a missing part. Are you up to adding/updating also the writing part? If not we can defer that to a follow-up PR.

tests/io/test_odim.py
```diff
@@ -177,6 +249,33 @@ def test_odim_dataset_has_close(odim_file):
...
+    # Build expected global attrs from raw /what/source.
+    expected = {}
+    with h5netcdf.File(filename, mode="r") as root:
+        if "what" in root:
+            source = root["what"].attrs.get("source", None)
+            print(f"Raw ODIM source attribute: {source}")
+            parsed = odim._parse_odim_source(source)
+            for odim_key, global_attr in odim._ODIM_SOURCE_TO_GLOBAL_ATTRS.items():
+                value = parsed.get(odim_key)
+                if value:
```
kmuehlbauer: This will skip "" (empty strings). Might this happen?

xradar/io/backends/odim.py
```diff
@@ -210,6 +210,50 @@ def _get_a1gate(where):
...
+    # We need to detect this and recover the original UTF-8 string if possible.
+    # e.g. "Sürgavere" is read as an ASCII string instead of UTF-8 and the invalid bytes (ü = \xc3\xbc) are encoded with
+    # surrogate escape, resulting in "S\udcc3\udcbcrgavere"
+    if any("\udc80" <= ch <= "\udcff" for ch in source):
+        try:
+            source = source.encode("utf-8", "surrogateescape").decode("utf-8")
+        except UnicodeError:
+            pass
+
+    # parse key-value pairs separated by commas, and then keys and values separated by colons
```
kmuehlbauer, line 240: Would be nice to refactor this into a helper, might get used in other parts of the library later.

xradar/io/backends/odim.py
```diff
@@ -766,7 +810,20 @@ def get_variables(self):
...
     def get_attrs(self):
-        attributes = {"Conventions": "ODIM_H5/V2_2"}
+        attributes = {"Conventions": "ODIM_H5/V2_2", "source": "radar"}
+
+        with self._manager.acquire_context(False) as root:
+            # try to extract additional global attributes from /what/source string, if available
+            # see _ODIM_SOURCE_TO_GLOBAL_ATTRS for mapping of ODIM keys to global attribute names
+            if "what" in root:
+                source = _maybe_decode(root["what"].attrs.get("source", None))
+                if source:
```
kmuehlbauer: There is a similar check in _parse_odim_source. Can this be simplified?

kmuehlbauer (changes requested):
@chfer Thanks, I think we can defer the writing part to later PR. I added a couple comments, otherwise LGTM!

tests/io/test_odim.py
chfer: Indeed if the value is a "" or 0, it would not be covered by the test, although it's present. I will strictly test on **value** being **None**.

```python
parsed = odim._parse_odim_source(source)
for odim_key, global_attr in odim._ODIM_SOURCE_TO_GLOBAL_ATTRS.items():
    value = parsed.get(odim_key)
    if value is not None:
        expected[global_attr] = value
```

xradar/io/backends/odim.py
chfer: If this is what you meant ... I created a helper function **_maybe_recover_surrogate** in common.py:

```python
def _maybe_recover_surrogate(txt):
    """Recover text that was decoded with surrogate escapes."""
    if not isinstance(txt, str):
        return txt

    if any("\udc80" <= ch <= "\udcff" for ch in txt):
        try:
            return txt.encode("utf-8", "surrogateescape").decode("utf-8")
        except UnicodeError:
            pass

    return txt

```

xradar/io/backends/odim.py
chfer: Indeed the extra check is not needed as it already has been done in the function **_maybe_decode**. I simplified this to :

```python
            if "what" in root:
                parsed = _parse_odim_source(root["what"].attrs.get("source"))
                for odim_key, global_attr in _ODIM_SOURCE_TO_GLOBAL_ATTRS.items():
                    value = parsed.get(odim_key)
                    if value is not None:
                        attributes[global_attr] = value
```

I also changed **"if value:"** to **"if value is not None:"** so that values like "", 0 and False are not skipped.

xradar/io/backends/odim.py
kmuehlbauer: @chfer That's great, too! Keep it! But I was referring to the string parser below.

chfer:
Hi @kmuehlbauer, I’ve addressed the comments in the latest commit. When you have a moment, would you mind taking another look? Thanks again!

kmuehlbauer (approved):
LGTM! Another maintainer need to merge, since I'll not be able to follow up due to traveling.

syedhamidali:
@kmuehlbauer Should I merge it?

PR #386, opened 2026-06-08
Towards default nominal angles

kmuehlbauer:
- [x] Possible successor of #383 
- [x] Tests added

### DIFF
--- a/tests/io/test_odim.py
+++ b/tests/io/test_odim.py
@@ -57,6 +57,126 @@ def test_get_azimuth_where(nrays):
     assert udiff[0] == 360.0 / nrays
 
 
+def test_odim_azimuth_nominal(odim_file):
+    import xarray as xr
+
+    ds = xr.open_dataset(odim_file, engine="odim", group="sweep_0")
+    azimuth = ds.azimuth.values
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@egouden I've taken your #383 and tweaked it toward my suggestion.

When you find the time, please have a look if this would also work from your perspective.

egouden:
Thanks @kmuehlbauer for handling this issue.

This is a more elaborated solution. It remains quite simple. I like it.

Is it the plan to tackle each format separately? I think for rainbow the nominal angle is included in the metadata.

kmuehlbauer:
@egouden The only thing I'm still not quite sure:

If we have a missing and a duplicate ray for some reason, like 0, 0, 1,..., 357, 359, it would create the nominal angles 0, 1, ..., 358, 359. Finally this would return an array which is one off for most of the sweep. I'll ponder about this a bit more, maybe we can get this right, too.

About the nominal angles: if formats support that, we should implement it. Otherwise we still could get nominal angles by auto extraction/reindex_dict. 

Would be nice to have conforming behaviour throughout the readers, though.

PR #387, opened 2026-06-09
MNT: Refactor CfRadial1

### LINKED ISSUE #379: Refactor CfRadial1 writer code quality before v1.0
The current CfRadial1 writer works in important cases, but the code quality should be improved before xradar v1.0. The main issues are unclear names, inconsistent structure, weak error handling, and some mismatch with the cleaner structure of the CfRadial1 backend reader.

Problems to address

* Helper names such as _main_info_mapper, _variable_mapper, and _calib_mapper are too generic.
* Local names such as data, dataset, result_dataset, vol_info, and calibs make the code harder to follow.
* calibs is used both as a boolean argument and later as a dataset variable:

syedhamidali:
- [x] Closes #379 
- [x] Tests added

Cleans up the CfRadial1 writer ahead of v1.0: descriptive helper names;
`_extract_root_dataset`
`_combine_sweeps`
`_map_radar_calibration`
`_collect_sweep_metadata`
 clearer locals/named constants, fixes the calibs bool/variable collision, more robust error handling (history attr, missing elevation, dtree=None), and a shared _build_cfradial1_dataset reused by transform.to_cfradial1 to drop ~40 duplicated lines. No behavior change; all CfRadial1/transform tests pass.

### DIFF
--- a/xradar/io/export/cfradial1.py
+++ b/xradar/io/export/cfradial1.py
@@ -27,10 +27,35 @@ __all__ = [
 
 __doc__ = __doc__.format("\n   ".join(__all__))
 
+import warnings
 from importlib.metadata import version
 
 import numpy as np
 import xarray as xr
+
+#: Per-sweep metadata variables that CfRadial1 stores as scalars (one value
+#: per sweep) rather than along a ray dimension.
+SWEEP_METADATA_VARS = (
+    "sweep_number",
+    "sweep_mode",
+    "polarization_mode",
+    "prt_mode",
+    "follow_mode",
+    "sweep_fixed_angle",
+    "sweep_start_ray_index",
+    "sweep_end_ray_index",
+)
+
+#: Subset of :data:`SWEEP_METADATA_VARS` gathered into the ``sweep``-indexed
+#: info dataset; the ray-index variables are computed separately.
+SWEEP_INFO_VARS = (
+    "sweep_number",
+    "sweep_mode",
+    "polarization_mode",
+    "prt_mode",
+    "follow_mode",
+    "sweep_fixed_angle",
+)
 
 
 def _first_valid_scalar(data_array):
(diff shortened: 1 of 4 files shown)


xradar/io/export/cfradial1.py
```diff
@@ -193,29 +212,28 @@ def _variable_mapper(dtree, dim0=None):
         combine_attrs="no_conflicts",
     )
 
-    drop_variables = [
+    per_sweep_vars = [
```
kmuehlbauer: This looks very similar to ``SWEEP_INFO_VARS``, only ``polarization_mode`` is missing. Is it the same?

xradar/io/export/cfradial1.py
```diff
@@ -225,53 +243,32 @@ def _sweep_info_mapper(dtree):
...
-        "prt_mode",
-        "follow_mode",
-        "sweep_fixed_angle",
-        "sweep_start_ray_index",
-        "sweep_end_ray_index",
-    ]
-
-    for var_name in sweep_vars:
-        var_data_list = [
+    for var_name in SWEEP_INFO_VARS:
```
kmuehlbauer, line 245: I assume ``sweep_start_ray_index`` does not need to be treated at this point. So, was it handled before?

xradar/io/export/cfradial1.py
```diff
@@ -327,53 +324,83 @@ def calculate_sweep_indices(dtree, dataset=None):
...
+    if "sweep_mode" in cfradial1_ds.variables:
+        cfradial1_ds["sweep_mode"] = cfradial1_ds["sweep_mode"].astype("S")
 
     # Update global attributes
-    dataset.attrs = dtree.attrs
-    dataset.attrs["Conventions"] = "Cf/Radial"
-    dataset.attrs["version"] = "1.2"
+    cfradial1_ds.attrs = dict(dtree.attrs)
+    cfradial1_ds.attrs["Conventions"] = "Cf/Radial"
+    cfradial1_ds.attrs["version"] = "1.2"
```
kmuehlbauer, line 368: OK, this was hard wired before. @syedhamidali would users want to export into other versions?

kmuehlbauer (commented):
@syedhamidali Thanks for taking this on. This is really nice improvement, not only in readability. I've added a couple of comments.

xradar/io/export/cfradial1.py
syedhamidali: @kmuehlbauer What should we do here? I would appreciate your thoughts on it.

kmuehlbauer: Let's keep it hardwired at "1.2" for now. Please open an issue where we can discuss if other CfRadial1 versions (1.3/1.4, https://ncar.github.io/CfRadial/) should also be supported.

xradar/io/export/cfradial1.py
syedhamidali: Good catch — they are effectively the same list, `polarization_mode` was the only difference, and that difference was pre-existing (the old `drop_variables` had the same five). It made no difference to the output because `polarization_mode` survives the `drop_vars` as a scalar and is then overwritten by `combined.update(sweep_info)` with the `sweep`-dimensioned version anyway.

I've unified it to use `SWEEP_INFO_VARS` and verified the exported dataset is unchanged — `xr.testing.assert_identical` on the full CfRadial1 export of `cfrad.20080604_002217_000_SPOL_v36_SUR.nc` (9 sweeps) passes before vs after, with `polarization_mode` keeping dims `('sweep',)` in both.

xradar/io/export/cfradial1.py
syedhamidali: Your assumption is right, and yes — it *was* handled before: the old `_sweep_info_mapper` looped over all eight variables including `sweep_start_ray_index`/`sweep_end_ray_index`. But that work was redundant, because the next step recomputes and overwrites both:

```python
combined.update(calculate_sweep_indices(dtree, combined))
```

So the exported values have always come from `calculate_sweep_indices`, never from the source tree. `SWEEP_INFO_VARS` therefore only gathers the six that aren't recomputed.

I verified this is output-preserving: exporting the same file with the pre-refactor writer (`79ba495`) and with this branch gives datasets that pass `xr.testing.assert_identical`, with the ray indices matching exactly.

xradar/io/export/cfradial1.py
syedhamidali: Sounds good — kept it hardwired at `"1.2"` here. Opened #388 to discuss supporting other CfRadial1 versions (1.3/1.4).

kmuehlbauer:
Thanks @syedhamidali! :rocket:

